PR #1, opened 2015-03-24
allow facility_code to be longer

ewheeler:
16 chars seems pretty small. 

in many use cases, there may not be a formal coding system for enumerating facilites -- with a longer char limit on this field, tracpro can accommodate facility names in the `facility_code` field

### DIFF
--- a/tracpro/contacts/migrations/0004_auto_20150324_1024.py
+++ b/tracpro/contacts/migrations/0004_auto_20150324_1024.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contacts', '0003_auto_20150224_1245'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='contact',
+            name='facility_code',
+            field=models.CharField(help_text='Facility code for this contact', max_length=160, null=True, verbose_name='Facility Code', blank=True),
+            preserve_default=True,
+        ),
+    ]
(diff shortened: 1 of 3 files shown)


nicpottier:
This looks fine for me, but @rowanseymour is out this week and I just want to double check with him first to see if there are any other assumptions at work here.

Are you ok waiting until next week to merge this?

ewheeler:
@nicpottier sounds good. i can use my fork for some demoing/testing until @rowanseymour is back

rowanseymour:
My understanding of this field was as a temporary holder for facility codes until a time when TracPro maintains a list of facility objects. In that case each facility object would have a code that can be matched with the code stored in the contact field on the RapidPro side, and on the TracPro side each contact just has a foreign key to their facility object.

If that is the case then, then overloading this field to sometimes contain facility names might make things trickier down the road. If not, then PR is fine.

ewheeler:
yes thats my understanding too. we're about to launch with 3000 schools and 10000 reporters in uganda. in this case, i think it makes more sense to punt on creating codes until we have a list of facilities that we can use to display a name based on the code. 

otherwise, users (like district education officers) won't be able to make sense of a new, arbitrary 16 char code and won't be able to really use tracpro effectively.

ewheeler:
I accidentally pushed two more commits here.. which will catch `decimal.InvalidOperation` exceptions raised when attempting to cast an empty string as a `Decimal`

Currently only `ValueError` exceptions are caught:
https://github.com/rapidpro/tracpro/blob/master/tracpro/polls/models.py#L514
https://github.com/rapidpro/tracpro/blob/master/tracpro/polls/models.py#L537

rowanseymour:
So what is the final decision about facilities? If this field is going to be used for facility names rather than codes then probably makes sense to rename it. Otherwise change looks fine.

The exception related commits also look fine.

ewheeler:
I agree that in the long-term this should be a code (with either additional support for facility objects in tracpro and/or integration with an external facility registry).

I'm happy to consider my facility names as 'long codes' for the time being...

PR #6, opened 2015-04-15
fetch 15 runs max at time to avoid rapidpro error 414

xkmato:


tracpro/polls/tasks.py
```diff
@@ -92,7 +92,9 @@ def fetch_org_updated_runs(org):
     incomplete_responses = Response.get_update_required(org)
 
     if incomplete_responses:
-        runs = client.get_runs(ids=[r.flow_run_id for r in incomplete_responses])
+        runs = []
+        for i in range(0, len(incomplete_responses), 15):
```
ewheeler: this looks like a great opportunity for `itertools`!
maybe something like:
```
for chunk in itertools.izip(*(iter(incomplete_responses),) * 15):
    runs += client.get_runs(ids=map(lambda x: x.flow_run_id, chunk)
```

xkmato: Yeah, I actually considered that but I realized handling the last list
would be complicated if it does not have 15 members. We will have to first
slice it off as well anyway, so I decided to go with simple slicing.

nicpottier: Why not just:

```python
runs = client.get_runs(ids=[r.flow_run_id for r in incomplete_responses[:15])
```

xkmato: That looks like it would ignore everything else and fetch the only first 15. If we are going with itertools then we will have to do something like this. Which is fine
```python
runs = client.get_runs(ids=map(lambda x: x.flow_run_id, incomplete_responses[:len(incomplete_responses)%15))
for chunk in itertools.izip(*(iter(incomplete_responses[len(incomplete_responses)%15:]),) * 15):
    runs += client.get_runs(ids=map(lambda x: x.flow_run_id, chunk)
```

ewheeler: @xkmato true, but this is a task that is repeated regularly. so with @nicpottier's version only 15 are synced each time the task is run. will eventually get them all

xkmato: But is that not a bit risky? Say if we have 16 responses every 5
minute(frequency of the task), then we will have about 12 delayed every
hour. I am not sure how that will impact the system but I would prefer not
to find out. What would be the impact of trying to get them all in one task?

ewheeler: good point. maybe:
```
while incomplete_responses:
    runs = client.get_runs(ids=[r.flow_run_id for r in incomplete_responses[:15])
```

nicpottier: I'm not actually familiar with this code, so will call in @rowanseymour for backup.

But ya, we should only be fetching up to where we DO have responses.

Other random note, this task strikes me as one that could back up (where the fetching could conceivably take longer than the time between task runs) so there should be a lock in here to prevent that.

rowanseymour: Agree with @xkmato's that this task doesn't run frequently enough to not fetch all incomplete runs.

Where did the 15 number come from? I'm wondering if we can't grab something like 50 at a time safely and so keep the number of API calls down. URL length limit is 2048 right?

And as @nicpottier points out - this task needs a lock so if it's still running when it's invoked again, it should do nothing.

xkmato: Yep it is definitely safe to bump the number up. I think 15 was a result of pure laziness. And implementing a lock is a great idea

rowanseymour: Unit test would also be good to verify that `client.get_runs` is called in batches

rowanseymour: Since we do these sort of batching operations quite a bit Im going to propose this as a re-usable solution: https://github.com/rapidpro/dash/pull/19. In this case you'd write...

```python
runs = []
for resp_chunk in chunks(list(incomplete_responses), 50):
  runs += client.get_runs(ids=[r.flow_run_id for r in resp_chunk])
```

ewheeler:
@xkmato can you pls update to use rowan's new `chunks` util method? also unit tests!

xkmato:
Updated... @rowanseymour already did unit test for chunks. Remember to upgrade when deploying

rowanseymour:
@xkmato this PR includes copies of all commits since Mar 24. I think I can merge it ok but are you merging against master? Rebasing?

rowanseymour:
@xkmato I merged and the history now has duplicate commits. I think in future you just have to make sure you merge in latest rapidpro/master before creating a PR. 

Also regrading the dash update - you should update pip-freeze.txt when a dependency changes and use this file for deploying since it has known (and in theory tested) versions of all dependencies. To update do...

```
pip freeze > pip-freeze.txt
```

You should then manually edit pip-freeze.txt revert the egg names of src dependencies back to what they were as pip can generate egg names with slashes that can't be folder names.

ewheeler:
thanks @xkmato and @rowanseymour 
i bumped our installation and the `414 Client Error: Request-URI Too Large` have stopped!

PR #15, opened 2015-08-18
Various changes

rebecca-caktus:
The non-relevant commits from upcoming work for hierarchical regions, split out in order to keep the signal-to-noise ratio of that work in control.


tracpro/test/cases.py
```diff
@@ -0,0 +1,155 @@
...
+                                 hour, minute, second, microsecond, tz)
+
+    def assertLoginRedirect(self, response, subdomain, next):
+        self.assertRedirects(
+            response,
+            'http://%s.testserver/users/login/?next=%s' % (subdomain, next))
+
+
+class TracProDataTest(TracProTest):
+    """Common data set-up."""
```
emullaney: Splitting this out makes sense. I extend TracProTest a few places, so I'll have to start using this new class.

emullaney:
Looks good :boat:

PR #17, opened 2015-08-18
Baseline chart model crudl

emullaney:
This is the Model and CRUDL for the new BaselineTerm object and new baseline app.


tracpro/static/libs/django-smart-selects/LICENSE.txt
```diff
@@ -0,0 +1,28 @@
+Copyright (c) 2009, Patrick Lauber
```
rebecca-caktus: Hmm, is this file needed?

tracpro/templates/frame.haml
```diff
@@ -39,6 +39,10 @@
     - block extra-less
     - block extra-style
 
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.11.4/jquery-ui.min.js"></script>
```
rebecca-caktus: Should jquery ui need to be loaded on every page? I think we're only using it on a couple of them.

tracpro/static/libs/django-smart-selects/LICENSE.txt
emullaney: It was unclear to me, but re-reading the license now, maybe not...

tracpro/baseline/migrations/0005_auto_20150817_1547.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
rebecca-caktus: I think that these 5 migrations can be squashed into one initial migration. 

(I would do it by deleting all `baseline` migrations and re-running `django-admin makemigrations baseline` to create a brand new initial migration. However [squashmigrations](https://docs.djangoproject.com/en/1.8/topics/migrations/#squashing-migrations) looks like it does essentially the same thing.)

tracpro/baseline/forms.py
```diff
@@ -0,0 +1,55 @@
...
+            self.fields['baseline_poll'].queryset = polls
+            self.fields['follow_up_poll'].queryset = polls
+
+    def clean(self, *args, **kwargs):
+        cleaned_data = super(BaselineTermForm, self).clean()
+
+        start_date = cleaned_data.get("start_date")
+        end_date = cleaned_data.get("end_date")
+
+        if start_date > end_date:
```
rebecca-caktus: It is possible that one or both of `start_date` or `end_date` could be `None`. `None > None` works but you get a `TypeError` if you try to compare a `datetime` with None.

You can change the check to `start_date and end_date and start_date > end_date` to cover the error condition while allowing the default `This field is required` messages to suffice if one or both fields is `None`.

tracpro/baseline/forms.py
```diff
@@ -0,0 +1,55 @@
+from django import forms
+
+from tracpro.polls.models import Poll
+from .models import BaselineTerm
+
+
+class BaselineTermForm(forms.ModelForm):
+    """
+    Form for Baseline Term
+    """
```
rebecca-caktus: After our talk yesterday, I'm not sure whether `region` will remain in this form. But if it does (here or elsewhere), `django-mptt` has some useful tools for forms: http://django-mptt.github.io/django-mptt/forms.html

rebecca-caktus: Would it be difficult to use javascript to hide the `Baseline Question` and `Follow up Question` fields until the corresponding `Poll` field has data?

rebecca-caktus:
Looks great so far!

tracpro/templates/frame.haml
emullaney: I've moved it out of frame into baseline templates.

tracpro/baseline/forms.py
emullaney: I've made this adjustment, please take a look at static\js\baseline.js and test it out.

emullaney: Also, I've remove region from the model for now.

tracpro/baseline/forms.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if start_date and end_date and start_date > end_date:
+            raise forms.ValidationError(
+                "Start date should be before end date."
+            )
+
+        baseline_question = cleaned_data.get("baseline_question")
+        follow_up_question = cleaned_data.get("follow_up_question")
+
+        if baseline_question and follow_up_question and baseline_question.pk == follow_up_question.pk:
```
rebecca-caktus: Django's [model equality implementation](https://github.com/django/django/blob/1.8.4/django/db/models/base.py#L506) essentially compares pks - so you can use `baseline_question == follow_up_question` here.

tracpro/static/js/baseline.js
```diff
@@ -0,0 +1,29 @@
+$(function() {
+        $("input[class^='datepicker']").datepicker();
```
rebecca-caktus: All but the first line are tabbed over once too many.

tracpro/static/js/baseline.js
```diff
@@ -0,0 +1,29 @@
...
+          if(this.value) {
+            $("#id_baseline_question").parent().parent().show();
+          }
+          else {
+            $("#id_baseline_question").parent().parent().hide();
+          }
+        });
+
+        // Display/hide question if poll is selected
+        if (!$("#id_follow_up_poll").val()) {
```
rebecca-caktus: You can remove some duplicated logic by calling `$("#id_follow_up_poll,#id_baseline_poll").change()` after hooking up the `'change'` behavior.

tracpro/static/js/baseline.js
```diff
@@ -0,0 +1,29 @@
...
+            $("#id_follow_up_question").parent().parent().hide();
+        }
+        $("#id_follow_up_poll").on("change", function() {
+          if(this.value) {
```
rebecca-caktus: This should be `$(this).val()`.

tracpro/static/js/baseline.js
```diff
@@ -0,0 +1,29 @@
...
+            $("#id_follow_up_question").parent().parent().show();
```
rebecca-caktus: What do you think about using `.closest(".form-group")`?

tracpro/templates/frame.haml
```diff
@@ -114,22 +116,31 @@
                         %a{ href:"{% url 'msgs.message_list' %}" }
                           - trans "Message Log"
 
+              %li.dropdown
```
rebecca-caktus: I see the "Management" menu when I am logged out, but it is empty.

rebecca-caktus:
Nifty `django-smart-selects` feature: If the poll only has one question, it is auto-selected!

tracpro/static/js/baseline.js
emullaney: Way better... thanks!

tracpro/static/js/baseline.js
emullaney: Awesome idea.

rebecca-caktus:
One comment then :zap:

PR #19, opened 2015-08-19
haml --> html transition

emullaney:
haml->html 

The reasons to remove haml from this project are two-fold:
- make this code more easily re-usable/approachable for other developers 
- remove one more thing that prevents us from porting to python 3


rebecca-caktus:
:fireworks:

emullaney:
@rebecca-caktus I'm going to close this PR and create a new one based on the latest version of master shortly.

PR #21, opened 2015-08-25
Translating haml->html

emullaney:
haml->html

The reasons to remove haml from this project are two-fold:

- make this code more easily re-usable/approachable for other developers
- remove one more thing that prevents us from porting to python 3


rebecca-caktus:
:pig:

PR #22, opened 2015-08-26
Reworked run fetching task for latest RapidPro changes 

rowanseymour:
which return runs using modified_on rather than created_on. Also added lock to prevent task from overlapping.

### DIFF
--- a/tracpro/orgs_ext/views.py
+++ b/tracpro/orgs_ext/views.py
@@ -55,7 +55,10 @@ class OrgExtCRUDL(OrgCRUDL):
         def get_last_flow_run_fetch(self, obj):
             result = obj.get_task_result(constants.TaskType.fetch_runs)
             if result:
-                return format_datetime(ms_to_datetime(result['time']))
+                return "%s (%d fetched)" % (
+                    format_datetime(ms_to_datetime(result['time'])),
+                    result.get('counts', {}).get('fetched', 0)
+                )
             else:
                 return None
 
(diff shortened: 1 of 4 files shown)


tracpro/orgs_ext/views.py
```diff
@@ -55,7 +55,10 @@ def get_last_contact_sync(self, obj):
         def get_last_flow_run_fetch(self, obj):
             result = obj.get_task_result(constants.TaskType.fetch_runs)
             if result:
-                return format_datetime(ms_to_datetime(result['time']))
+                return "%s (%d fetched)" % (
```
rowanseymour: Just adds the number of fetched runs to the org view, similar to what there is for contact sync, e.g.

__Last Flow Run Fetch__ Aug 26, 2015 13:16 (0 fetched)

tracpro/polls/tasks.py
```diff
@@ -20,102 +21,59 @@ def fetch_all_runs():
...
-            runs = runs[0:-1]
-
-    logger.info("Fetched %d new runs for org #%d (after=%s)"
-                % (len(runs), org.id, format_iso8601(last_time) if last_time else 'Never'))
-
-    if runs:
-        newest_run = runs[0]
+    total_runs = 0
+    for poll in Poll.get_all(org):
+        poll_runs = client.get_runs(flows=[poll.flow_uuid], after=last_time, before=until)
```
rowanseymour: It's very likely at some point in the future, for performance reasons, we'll have to restrict the API to fetching runs by a single flow, or for all org flows... so future proofing this now.

rebecca-caktus:
:+1:

PR #23, opened 2015-09-01
Handle region deactivation

rebecca-caktus:
When a region is deactivated, its children become the children of its parent.


rebecca-caktus:
This is ready to review. I rebased the branch so if you have a local copy you'll want to update with something like `git remote update && git checkout region-deactivation && git reset --hard @{upstream}`.

The build is failing only due to low test coverage. I don't know how test coverage dropped so dramatically given the limited scope of this request and the additional tests that are included. I'll figure it out later, but the pull request should otherwise be in good shape.

rebecca-caktus:
Oh, I see what is decreasing the coverage. I added the `Contact` factory in `contacts/tests/`, which overshadows `contacts/tests.py`. This will be fixed by #25.

rebecca-caktus:
Test coverage is now restored.

emullaney:
great test coverage! :grapes:

PR #24, opened 2015-09-01
Baseline chart functionality and spoofing

emullaney:
DISPLAYING/CREATING CHARTS
- View baseline charts on the Baseline Read page
- Redirect the baseline chart updates/creates back to the Read page for quicker viewing of charts.
- All logic behind baseline charting is contained in the model and view with the chart javascript in the template

SPOOFED DATA
- New ability to create spoof data which requires a model change on polls.models.Response.flow_run_id which can now be NULL for this spoofed data
- TODO: Figure out a way to clear out spoofed data and specify spoofed data in the charts themselves so as not to confuse it with real data


tracpro/baseline/forms.py
```diff
@@ -44,10 +45,82 @@ def clean(self, *args, **kwargs):
...
+        help_text=_("Follow up answers will be created for each contact within the minimum/maximum range."))
+    follow_up_maximum = forms.IntegerField(
+        help_text=_("Follow up answers will be created for each contact within the minimum/maximum range."))
+
+    def __init__(self, *args, **kwargs):
+        org = kwargs.pop('org')
+
+        super(SpoofDataForm, self).__init__(*args, **kwargs)
+
+        if org:
```
rebecca-caktus: Looks like this argument comes from `request.org`. Is it possible to get to this page without an org? If not, the conditional can be removed.

rebecca-caktus: Ditto for the `BaselineTermForm`.

tracpro/baseline/tests/test_models.py
```diff
@@ -0,0 +1,142 @@
+from datetime import datetime
+from dateutil import rrule
+
+from tracpro.test.cases import TracProDataTest
+from ..models import BaselineTerm
```
rebecca-caktus: Relative imports should be separated from the other imports.

tracpro/baseline/views.py
```diff
@@ -6,17 +12,22 @@
     SmartCRUDL, SmartCreateView, SmartDeleteView,
     SmartListView, SmartReadView, SmartUpdateView
 )
+from smartmin.users.views import SmartFormView
 
 from .models import BaselineTerm
-from .forms import BaselineTermForm
+from .forms import BaselineTermForm, SpoofDataForm
+from tracpro.polls.models import (
```
rebecca-caktus: Relative imports usually go last.

tracpro/templates/baseline/baselineterm_read.html
```diff
@@ -3,9 +3,70 @@
...
\ No newline at end of file
+{% block read-buttons %}{% endblock read-buttons %}
+
+{% block extra-script %}
+<script src="http://code.highcharts.com/modules/exporting.js"></script>
+<script>
+$(function () {
+    $('#container').highcharts({
+        title: {
+            text: '{{ baselineterm.name }}'
```
rebecca-caktus: If the baseline term name has an apostrophe in it then this could get ugly. You could [escape such values](https://docs.djangoproject.com/en/dev/ref/templates/builtins/#escapejs), but I'm wondering: what do you think about creating a JSON endpoint from which to grab this data?

tracpro/templates/baseline/baselineterm_read.html
```diff
@@ -3,9 +3,70 @@
...
+        xAxis: {
+            categories: [
+                        {% for date in date_list %}
+                            '{{ date }}',
+                        {% endfor %}
+                        ]
+        },
+        yAxis: {
+            title: {
+                text: '' //TODO: Figure out what the yAxis header should be
```
rebecca-caktus: Maybe this text can be configurable per object?

tracpro/templates/baseline/baselineterm_data.html
```diff
@@ -0,0 +1,147 @@
...
+    {% endif %}
+
+    {% if form.non_field_errors %}
+        <ul class="field-errors">
+            <li>{{ form.non_field_errors }}</li>
+        </ul>
+    {% endif %}
+
+    <div>
+      <form id="spoofform" class="smartmin-form form-horizontal ng-pristine ng-valid" action="." method="post">
```
rebecca-caktus: Smartmin has some cool tools for displaying bootstrap forms. This template powers the create and update views: https://github.com/nyaruka/smartmin/blob/master/smartmin/templates/smartmin/form.html

rebecca-caktus:
I am wondering if it is possible to either 1) keep track of whether data is 'spoofed', such that it can be deleted later, or 2) make the data ephemeral such that we don't have to worry about it mucking up real data. The latter might require more thought than its worth.

PR #25, opened 2015-09-02
Test directory structure

rebecca-caktus:
* Use a `tests/` directory structure rather than `tests.py`.
* Limits line length in some files.
* Keep all forms in `forms.py` files.


emullaney:
looks great! :banana:

PR #26, opened 2015-09-04
WIP - Hierarchical Region Permissions

rebecca-caktus:
Contains scaffolding for hierarchical region permissions: 

* Display drop-down region menu as a hierarchy with active regions highlighted
* Manage a session variable called `include_subregions` to create a site-wide flag for whether sub-regions should be included.
* Update user region utilities to allow checking for direct regions and sub-regions

Integration throughout the app is in progress:

* [x] Contacts
* [x] Inbox messages
* [x] Messages
* [x] Home page
* [ ] Baseline charts
* [x] Polls
* [x] Profiles


rebecca-caktus:
Here's where we stand now:
 
* Existing tests pass.
* `baseline` app not yet fully updated; work in progress.
* Will harden with tests as the week continues.

tracpro/profiles/__init__.py
```diff
@@ -52,18 +53,31 @@ def _user_get_full_name(user):
...
-    return get_obj_cacheable(user, '_regions', calculate)
+
+def _user_get_all_regions(user, org):
+    """Return org regions user has direct or implied (by hierarchy) permission for."""
+    def calculate():
+        regions = user.get_direct_regions(org).get_descendants(include_self=True)
+        regions = regions.filter(is_active=True)
+        return regions
+    attr_name = '_regions_with_descendants_{}'.format(org.pk)  # cache per org
+    return get_obj_cacheable(user, attr_name, calculate)
```
emullaney: I'm all for making granular, re-usable methods but I'm wondering why you split out get_direct_regions and get_all_regions because get_all_regions() is the only one that seems to be called.

tracpro/templates/frame.html
```diff
@@ -232,28 +235,58 @@
...
+                                {% csrf_token %}
+                                <input type="hidden" name="include_subregions"
+                                       value="{% if request.include_subregions %}0{% else %}1{% endif %}" />
+                                <input type="hidden" name="next"
+                                       value="{{ request.get_full_path }}" />
+                                <input type="submit" class="hidden" />
+                              </form>
+                              {% if request.include_subregions %}
+                                <a class="toggle-subregions" href="#"
+                                   title="Currently showing data for {{ request.region }} and all sub-regions. Click to show data from {{ request.region }} only.">
```
emullaney: Really like the title for this and the Show link, they give the user great feedback.

tracpro/groups/views.py
```diff
@@ -23,6 +26,42 @@
...
+            raise HttpResponseBadRequest(
+                "{} should be either '0' or '1'".format(
+                    self.post_param))
+
+        # Redirect to the next path.
+        # TODO: Double-check validation of `next` parameter.
+        next_path = request.POST.get(self.next_param)
+        return redirect(next_path)
+
+
```
emullaney: Clever stuff!

emullaney:
One comment see tracpro/profiles/__init__.py.
One issue I noticed when viewing Pollrun "Responses" tab via this link (which you will need to edit to test out your local data) -  http://caktus.localhost:8000/response/by_pollrun/325/?_region=1

... after those two things :shipit:

tracpro/profiles/__init__.py
rebecca-caktus: Good question! I wrote this method before I made most actual changes. I had thought I would need both of these methods in different places, but as you point out, so far that has not been the case.

PR #27, opened 2015-09-08
Spoof data form edits

emullaney:
Make spoof fields required. 
Override label on question drop down to include poll name.

### DIFF
--- a/tracpro/baseline/forms.py
+++ b/tracpro/baseline/forms.py
@@ -54,32 +54,45 @@ class BaselineTermForm(forms.ModelForm):
         return cleaned_data
 
 
+class CustomModelChoiceField(forms.ModelChoiceField):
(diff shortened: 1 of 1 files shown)


rebecca-caktus:
The `required=True` argument to the fields is not needed - form fields are required by default: https://docs.djangoproject.com/en/1.8/ref/forms/fields/#required

tracpro/baseline/forms.py
```diff
@@ -54,32 +54,45 @@ def clean(self, *args, **kwargs):
         return cleaned_data
 
 
+class CustomModelChoiceField(forms.ModelChoiceField):
```
rebecca-caktus: What do you think about calling this something more specific like `QuestionModelChoiceField`?

rebecca-caktus:
:rocket:

PR #28, opened 2015-09-08
Baseline data values: include all baselines across dates

emullaney:
~~Select baseline data based on least recent answer per contact.~~
Select baseline data points across all dates.
Update tests for this functionality.


tracpro/baseline/models.py
```diff
@@ -32,7 +32,7 @@ class BaselineTerm(models.Model):
         chained_model_field='poll',
         auto_choose=True,
         related_name="baseline_terms",
-        help_text=_("The most recent response per user will be used as the baseline.")
+        help_text=_("The least recent response per user will be used as the baseline.")
```
rebecca-caktus: Hmm, I don't think this is as clear in the reverse. What do you think about: "The first response in the time period per user..."

rebecca-caktus:
Love how simple this change was. 

One silly `flake8` error, one comment, then good to :goat:

emullaney:
This one is going to be a bit more complex to allow for multiple baselines over time. Going to re-think the wording and re-work the logic on this branch.

tracpro/baseline/forms.py
```diff
@@ -68,7 +68,9 @@ class SpoofDataForm(forms.Form):
     start_date = forms.DateField(help_text=_(
         "Baseline poll data will be submitted on this date. " +
         "Follow up data will start on this date."))
-    end_date = forms.DateField(help_text=_("Follow up data will end on this date."))
+    end_date = forms.DateField(help_text=_(
+        "Follow up data will end on this date. " +
```
rebecca-caktus: `+` signs are unnecessary to concatenate strings across lines when in `()`.

tracpro/baseline/models.py
```diff
@@ -33,7 +33,7 @@ class BaselineTerm(models.Model):
...
```
rebecca-caktus: I think this help text needs to be updated.

tracpro/baseline/views.py
```diff
@@ -144,8 +155,27 @@ def form_valid(self, form):
...
+            end = self.form.cleaned_data['end_date']
+
+            # Create a single PollRun for the Baseline Poll
+            self.create_baseline(baseline_question.poll, start, contacts, baseline_question,
+                                 baseline_minimum, baseline_maximum)
+
             # Create a PollRun for each date from start to end dates for the Follow Up Poll
+            loop_count = 1
             for follow_up_date in rrule.rrule(rrule.DAILY, dtstart=start, until=end):
+                if loop_count == 7:  # On the seventh day, add another set of baseline data
```
rebecca-caktus: Hmm, if my finger-counting is correct I think this is one-off.

A possible tidy solution: 

```
for loop_count, follow_up_date in enumerate(rrule.rrule(rrule.DAILY, dtstart=start, until=end)):
    if loop_count % 7 == 0:
        ...
```

tracpro/baseline/views.py
```diff
@@ -74,18 +74,40 @@ def get_context_data(self, **kwargs):
 
             region = self.request.region
 
-            baseline_dict = self.object.get_baseline(region=region)
-            context['baseline_dict'] = baseline_dict
+            baseline_dict, baseline_dates = self.object.get_baseline(region=region)
 
             follow_ups, dates = self.object.get_follow_up(region=region)
 
+            """
```
rebecca-caktus: I'm not sure that I have ever seen block comments anywhere except for a docstring, but I don't have any good sources telling me that's the way it should be. Do you have a reason or reference for doing this?

emullaney: If that's not the norm in python, I will go back to single # comments. No worries.

tracpro/baseline/views.py
emullaney: Cool stuff. Will re-do this. Forgot about enumerate().

rebecca-caktus:
I see these warnings fly by when I create spoofed data:

```
RuntimeWarning: DateTimeField Response.updated_on received a naive datetime (2016-01-01 14:27:27.803316) while time zone support is active.
```

tracpro/baseline/views.py
```diff
@@ -144,9 +150,27 @@ def form_valid(self, form):
...
+            contacts = self.form.cleaned_data['contacts']
+            baseline_minimum = self.form.cleaned_data['baseline_minimum']
+            baseline_maximum = self.form.cleaned_data['baseline_maximum']
+            follow_up_minimum = self.form.cleaned_data['follow_up_minimum']
+            follow_up_maximum = self.form.cleaned_data['follow_up_maximum']
+            start = self.form.cleaned_data['start_date']
+            end = self.form.cleaned_data['end_date']
+
+            # Create a single PollRun for the Baseline Poll
+            self.create_baseline(baseline_question.poll, start, contacts, baseline_question,
```
rebecca-caktus: It looks like this could be moved into the for loop if you remove the `loop_count > 0` conditional.

emullaney: Goooood point....

rebecca-caktus:
:+1:

PR #30, opened 2015-09-09
Hierarchical region permissions in `baseline` app

rebecca-caktus:
This also fixes a small bug wherein `PollRun.pollrun_type` is not being set on spoof data.

When merged, we'll want to set the `pollrun_type` of such items:

```
from tracpro.polls.models import PollRun
PollRun.objects.filter(region=None, pollrun_type="").update(pollrun_type=PollRun.TYPE_UNIVERSAL)
```

### DIFF
--- a/tracpro/baseline/models.py
+++ b/tracpro/baseline/models.py
@@ -1,4 +1,7 @@
+from collections import defaultdict
+
 from django.db import models
+from django.db.models import F
 from django.utils.translation import ugettext_lazy as _
 
 from smart_selects.db_fields import ChainedForeignKey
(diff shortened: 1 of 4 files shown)


rebecca-caktus:
As it turns out, this fixes a small bug wherein `PollRun.pollrun_type` is not being set on spoof data.

When merged, we'll want to set the `pollrun_type` of such items:

```
from tracpro.polls.models import PollRun
PollRun.objects.filter(region=None, pollrun_type="").update(pollrun_type=PollRun.TYPE_UNIVERSAL)
```

emullaney:
:100:

PR #33, opened 2015-09-10
Baseline modifications

rebecca-caktus:
* Introduce a new `pollrun_type`, `TYPE_SPOOFED`. The previous pull request uses `get_or_create_universal` to create spoofed `PollRun` instances. This caused a bug where valid universal PollRuns were deleted when removing spoofed data.
* Convert `BaselineTerm.start_date` and `BaselineTerm.end_date` to store dates rather than datetimes. I do not expect this to cause loss of data.
* Factor out common pattern in retrieval of `BaselineTerm` answers.
* Fix missing baseline data line (cannot iterate through `defaultdict` in a Django template).

### DIFF
--- a/tracpro/baseline/migrations/0004_auto_20150910_2018.py
+++ b/tracpro/baseline/migrations/0004_auto_20150910_2018.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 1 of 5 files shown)


rebecca-caktus:
Still trying to figure out why the "baseline" data line is not appearing on baseline charts (but "follow up" is).

emullaney:
Great catch on defaultdict. :cactus:

PR #34, opened 2015-09-11
Filter data on baseline chart page by region.

emullaney:
New drop-down on the baseline read page that allows user to filter chart by region. 

Also includes a filter by the regions from the session variable.

(This is take 2 after the first attempt got out of sync with master.)

### DIFF
--- a/tracpro/baseline/models.py
+++ b/tracpro/baseline/models.py
@@ -55,7 +55,7 @@ class BaselineTerm(models.Model):
         baseline_terms = cls.objects.filter(org=org)
         return baseline_terms
 
-    def _get_answers(self, question, regions):
+    def _get_answers(self, question, regions, region_selected):
         """
         Retrieve answers to the question that are relevant for this
         BaselineTerm.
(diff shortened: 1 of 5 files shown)


tracpro/baseline/models.py
```diff
@@ -68,15 +68,23 @@ def _get_answers(self, question, regions):
...
+                                     'region_name').distinct('response__contact__region')
+
         if regions:
             answers = answers.filter(response__contact__region__in=regions)
-        return answers
+            all_regions = answers.values('response__contact__region',
+                                         'region_name').distinct('response__contact__region')
+
+        if region_selected:
+            answers = answers.filter(response__contact__region__in=[region_selected])
```
rebecca-caktus: More simple: `answers.filter(response__contact__region=region_selected)`.

tracpro/baseline/views.py
```diff
@@ -73,8 +75,17 @@ def derive_queryset(self, **kwargs):
         def get_context_data(self, **kwargs):
             context = super(BaselineTermCRUDL.Read, self).get_context_data(**kwargs)
 
-            baseline_dict, baseline_dates = self.object.get_baseline(self.request.data_regions)
-            follow_ups, dates = self.object.get_follow_up(self.request.data_regions)
+            # Get the region from the region filter drop-down, if it was selected
+            region = int(self.request.GET.get('region', 0))
```
rebecca-caktus: We should surround this with `try... except...` just in case a bad value is passed in.

tracpro/templates/baseline/baselineterm_header.html
```diff
@@ -4,26 +4,47 @@
...
+        {% for region in all_regions %}
+          {% for key, value in region.items %}
+              {% if key == "response__contact__region" %}
+                  {% if value == region_selected %}
+                    <option selected='selected' value='{{ value }}'>
+                  {% else %}
+                    <option value='{{ value }}'>
+                  {% endif %}
+              {% else %}
+                  {{ value }}</option>
```
rebecca-caktus: I'm having trouble seeing how this `</option>` can match up to an opening `<option>` tag.

tracpro/baseline/models.py
```diff
@@ -68,15 +68,23 @@ def _get_answers(self, question, regions):
...
         answers = answers.annotate(region_name=F('response__contact__region__name'))
         answers = answers.select_related('response', 'response__contact')
+
+        all_regions = answers.values('response__contact__region',
+                                     'region_name').distinct('response__contact__region')
+
         if regions:
             answers = answers.filter(response__contact__region__in=regions)
-        return answers
+            all_regions = answers.values('response__contact__region',
```
rebecca-caktus: The eventual use of this in the template is kind of confusing. What about changing this to something like: `all_regions = answers.values_list('response__contact__region', 'region_name')`

Then you could use something like this in the template: `{% for region_id, region_name in all_regions %}`

rebecca-caktus:
Another post-vacation :shipit:

PR #37, opened 2015-09-16
Updating to highcharts (js library) 4.1.8

emullaney:
This just replaces the highcharts library with the latest and greatest from highcharts.


rebecca-caktus:
I'm kind of fond of keeping library code in a versioned subdirectory of the [static/libs/](https://github.com/rapidpro/tracpro/tree/master/tracpro/static/libs) directory. I find it makes it easier to remember not to make changes to library files.

PR #38, opened 2015-09-16
Task Updates and Fix

emullaney:
- Adding exclude() to filter that gets the newest Response time to exclude out spoof data. 
- Reorder import statements to follow conventions.
- Rename r variable to longer less generic name for redis connection.


rebecca-caktus:
Change looks good!

This was a tricky issue to find. How possible would it be to add a test?

emullaney:
@rebecca-caktus not sure how to test a task that connects to the API... any recs?

rebecca-caktus:
Use `mock.patch` to set the results that you expect from the API. See `tracpro.groups.tests.test_models.TestRegion` for some examples. And let's look at it together if you have questions!

PR #41, opened 2015-09-17
Create subdomain whitelist

rebecca-caktus:
In lieu of allowing arbitrary subdomains (for now), this creates a whitelist that can be modified for staging or production.


emullaney:
:cookie:

PR #45, opened 2015-09-21
Fix baseline date data

emullaney:
Needed to adjust the logic to deal with displaying multiple sets of dates on the same chart. 

If a poll is given to multiple regions over different dates, and we want to display all region data on one chart with correct dates on x-axis.


rebecca-caktus:
:+1:

PR #35, opened 2015-09-11
Home page baseline chart

emullaney:
Display a baseline chart at the top of the homepage.
This chart is selected by looping through all baseline terms, in reverse order from the end date, until one is found with data. 
The chart is clickable and links to the chart's read page.
Charting function was moved into a utility on the baseline app, and chart javascript was moved into an include template.

PR #46, opened 2015-09-21
Post PR adjustments

emullaney:
Because @rebecca-caktus had a few outstanding comments for fixes from a couple weeks back, I went ahead and fixed them. 

Original comments from original PR's found here: https://github.com/rapidpro/tracpro/pull/34
and here: https://github.com/rapidpro/tracpro/pull/35

### DIFF
--- a/tracpro/baseline/views.py
+++ b/tracpro/baseline/views.py
@@ -77,7 +77,13 @@ class BaselineTermCRUDL(SmartCRUDL):
             context = super(BaselineTermCRUDL.Read, self).get_context_data(**kwargs)
 
             # Get the region from the region filter drop-down, if it was selected
-            region = int(self.request.GET.get('region', 0))
+            try:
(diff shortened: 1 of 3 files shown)


tracpro/baseline/views.py
```diff
@@ -77,7 +77,14 @@ def get_context_data(self, **kwargs):
             context = super(BaselineTermCRUDL.Read, self).get_context_data(**kwargs)
 
             # Get the region from the region filter drop-down, if it was selected
-            region = int(self.request.GET.get('region', 0))
+            try:
+                region = int(self.request.GET.get('region', 0))
+            except:
```
rebecca-caktus: Let's narrow this to `ValueError`.

Casting as an `int` can also lead to a `TypeError` if `None` is passed - but that shouldn't ever come into play here.

PR #47, opened 2015-09-22
Fix for clearing spoofed data, 

emullaney:
- Filter by current org only
- Filter by TYPE_SPOOFED 
- OrgPermsMixin for the view
- Improved "are you sure?" message


rebecca-caktus:
Code changes look good :balloon:

PR #48, opened 2015-09-23
Handle non-existant Poll in PollRun.by_poll

rebecca-caktus:
Fixes this error:

```
  File "/var/www/tracpro/source/tracpro/polls/views.py", line 318, in derive_queryset
    return self.derive_poll().get_pollruns(
  File "/var/www/tracpro/source/tracpro/polls/views.py", line 315, in derive_poll
    return get_obj_cacheable(self, '_poll', fetch)
  File "/var/www/tracpro/env/src/dash/dash/utils/__init__.py", line 85, in get_obj_cacheable
    calculated = calculate()
  File "/var/www/tracpro/source/tracpro/polls/views.py", line 314, in fetch
    pk=self.kwargs['poll'], org=self.request.org, is_active=True)
DoesNotExist: Poll matching query does not exist.
```

The Poll in question was "2", which doesn't look nefarious. I will investigate in case a user was erroneously being led to a Poll they should not be able to see.

### DIFF
--- a/tracpro/polls/models.py
+++ b/tracpro/polls/models.py
@@ -44,6 +44,15 @@ class Window(Enum):
             return (now - relativedelta(days=self.ordinal), now)
 
 
+class PollQuerySet(models.QuerySet):
+
+    def active(self):
+        return self.filter(is_active=True)
+
+    def by_org(self, org):
+        return self.filter(org=org)
+
+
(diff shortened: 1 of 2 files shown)


emullaney:
:sun_with_face:

emullaney:
mixins! :eggplant:

PR #49, opened 2015-09-23
Baseline data first results only

emullaney:
- Select the baseline results from the sum of the *first* set of regional contacts' results.
- Assume baseline will not change over time.
- Updated test to test for this.
- Updated spoof to create a single baseline at beginning of spoofed dataset.


rebecca-caktus:
:+1:

PR #50, opened 2015-09-23
Handle if response is not for a requested reporter group

rebecca-caktus:
Fixes this error:

```
  File "/var/www/tracpro/source/tracpro/polls/views.py", line 245, in get_context_data
    group = reporting_groups_by_id[group_id] if group_id else None
KeyError: 10
```


emullaney:
:dango:

PR #52, opened 2015-09-24
Baseline Total Part 2

emullaney:
Baseline total for all dates, so that the baseline does not change over time.


rebecca-caktus:
:anchor:

PR #53, opened 2015-09-24
Chart summary data

emullaney:
- New calculation: standard deviation and mean for baseline and follow ups
- New requirement: numpy (for calculations)

On the home page:
- Alter design to make the chart fill up more of the screen. 

On the chart page:
- Move all chart div styles into trac.less
- Increase chart size
- Move legend to right of chart
- Display new data (mean and st dev) in the legend


rebecca-caktus:
Could we reduce the precision of the summary data?

rebecca-caktus:
:boom:

PR #55, opened 2015-09-28
Small changes

rebecca-caktus:
A few changes that require migrations:

* CHOICES specified with definitive order, and migration to update. Prevents extra system-dependent migrations from being created.
* Added a line to the build to fail if required migrations are not present.
* Remove the backwards relation from polls/questions to the `BaselineTerm` model.
* Disallow null `y_axis_label` values to prevent "None" from showing up on the y axis if the label is not defined.


emullaney:
:peach:

emullaney:
:banana:

PR #58, opened 2015-09-29
Constant baseline

emullaney:
New form to allow user to chart a constant baseline value


rebecca-caktus:
I'll wait for the region redo to be merged & this branch updated before reviewing...

rebecca-caktus:
This looks good! I kind of envisioned having this on the configuration page for the BaselineTerm (as an alternative to `baseline_question`). It would probably save us some calculation time esp. as the data sets get larger. Maybe that's an option for the future. Functionally this looks great. :shipit:

PR #61, opened 2015-09-30
Region redo

emullaney:
All region data in one baseline and one follow up series on the chart.
Simplifying the chart code (by a lot!)

### DIFF
--- a/tracpro/baseline/models.py
+++ b/tracpro/baseline/models.py
@@ -53,7 +53,7 @@ class BaselineTerm(models.Model):
         baseline_terms = cls.objects.filter(org=org)
         return baseline_terms
 
-    def _get_answers(self, question, regions, region_selected):
+    def _get_answers(self, question, regions, region_selected=None):
         """
         Retrieve answers to the question that are relevant for this
         BaselineTerm.
@@ -80,7 +80,7 @@ class BaselineTerm(models.Model):
 
         return answers, all_regions
 
-    def get_baseline(self, regions, region_selected):
+    def get_baseline(self, regions, region_selected=None):
         """ Get all baseline responses """
         answers, all_regions = self._get_answers(self.baseline_question, regions, region_selected)
         # Retrieve the first result per contact for baseline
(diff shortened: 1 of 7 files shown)


tracpro/baseline/models.py
```diff
@@ -83,36 +83,20 @@ def _get_answers(self, question, regions, region_selected):
...
+        """ Get all follow up responses """
         answers, all_regions = self._get_answers(self.follow_up_question, regions, region_selected)
 
-        # Loop through all regions in answers and create
-        # a dict of values and dates per Region
-        # ex.
-        # { 'Kumpala': {'values': [35,...], 'dates': [datetime.date(2015, 8, 12),...]} }
-        region_answers = {}
-        dates_dict = {}
         dates = []
```
rebecca-caktus: `dates` is defined below so doesn't need to be defined here.

tracpro/baseline/models.py
```diff
@@ -83,36 +83,20 @@ def _get_answers(self, question, regions, region_selected):
...
-            answer_sums, dates = answers_by_region.numeric_sum_group_by_date()
-            region_answers[region_name] = {'values': answer_sums}
-            dates_dict[region_name] = dates
-        return region_answers, dates_dict, all_regions
+        answers = answers.order_by('submitted_on')
+        answers_list, dates = answers.numeric_sum_group_by_date()
+        return answers_list, dates, all_regions
 
     def check_for_data(self, regions):
         answers, all_regions = self._get_answers(self.baseline_question, regions, 0)
```
rebecca-caktus: This could be `answers.exists()`

tracpro/baseline/tests/test_models.py
```diff
@@ -86,26 +86,13 @@ def setUp(self):
...
-        self.assertEqual(len(baseline_dict), 2)  # Two regions, two sets of baseline values
-        # Two answers, 10 + 20 = 30
-        self.assertEqual(
-            baseline_dict[self.region1.name]["values"],
-            [30])
-        # One answer, 30 = 30
-        self.assertEqual(
-            baseline_dict[self.region2.name]["values"],
-            [30])
+        baseline_total, dates = self.baselineterm.get_baseline(regions=None, region_selected=0)
```
rebecca-caktus: I think that `None` rather than `0` would probably make more sense here (and a few other places).

rebecca-caktus: Alternatively adding `None` as the default for `region_selected` on the method itself.

rebecca-caktus:
Wow, code changes & charts both look great!

The few comments above are hardly blockers, just things I noticed. :shipit: at will!

PR #63, opened 2015-10-01
Response rate redux

emullaney:
Calculation for response rate.
Final item from https://trello.com/c/W8tBbv1V/47-feedback-include-summary-mean-avg-etc-on-graph-right-side.


rebecca-caktus:
:+1:

PR #65, opened 2015-10-05
Filter PollRun by org in addition to region

rebecca-caktus:
Fixed bug: If region is None, then we're returning results for other orgs as well.

Also renames `temba` -> `temba_client` as required by recent changes to rapidpro/rapidpro-python. (I've updated the master branches of our forks of smartmin, rapidpro-python, and dash with the latest from their upstream counterparts.)


emullaney:
fast fix! :+1:

PR #69, opened 2015-10-07
Updated Contact Sync

emullaney:
- This is completely reliant on a related update on dash repo: https://github.com/caktus/dash/tree/contact-sync
- This update will sync contacts based on those that have been updated in RapidPro only. 
- We now keep track of a new redis cached variable in order to filter out results by this date: LAST_FETCHED_CONTACTS_RUN_TIME_KEY


rebecca-caktus:
Some time passes between when we sync contacts, and when we calculate the last time that we synced as `timezone.now()`. This time period might be small, or it could be quite large if a lot of data is returned/processed. If a contact is updated during this time window, we won't find it the next time that sync is run.

According to the [docs](https://app.rapidpro.io/api/v1/contacts), RapidPro accepts both an `after` and `before` parameter. Making use of both of these parameters will eliminate the gap. This is an approximation of what I mean:

```
last_time_key = LAST_FETCHED_CONTACTS_RUN_TIME_KEY % org.pk
after = redis_connection.get(last_time_key)  # if None, then this will pull from the beginning of time
before = timezone.now()
sync_pull_contacts(..., after=after, before=before)
redis_connection.set(last_time_key, format_iso8601(before))
```

rebecca-caktus:
Worked out some issues with @emullaney in person - last time synced should be set to the most recent `TembaContact.modified_on` value.

emullaney:
This works perfectly when run from the command line. For some reason, when run from celery it's getting stuck.

Any suggestions, @rebecca-caktus?

tracpro/contacts/migrations/0006_contact_temba_modified_on_preset_data.py
```diff
@@ -0,0 +1,41 @@
...
+
+from django.db import models, migrations
+
+from temba_client.base import TembaAPIError
+from dash.orgs.models import Org
+
+from ..models import Contact
+
+
+def no_op(apps, schema_editor):
```
rebecca-caktus: New in Django 1.8: https://docs.djangoproject.com/en/1.8/ref/migration-operations/#django.db.migrations.operations.RunPython.noop

rebecca-caktus:
I don't know if the data migration is necessary. Everything should be updated magically when `sync_all_contacts` is run for the first time. I'm afraid of potential data loss in the in-between periods.

rebecca-caktus:
This works great for me locally. Let's try it out on staging to see if there are any celery issues.

rebecca-caktus:
:shipit:

PR #70, opened 2015-10-07
Homepage edits

emullaney:
- poll title in chart header
- moving legend to top of the chart
- poll start date and end date on the bottom, under the x-axis
- rename latest baseline table; remove start date, baseline question and follow-up question.
- increase size of chart


rebecca-caktus:
:shipit:

PR #72, opened 2015-10-13
Rename Baseline -> Indicators, Follow Up -> Observations

rebecca-caktus:
On the front end only.


emullaney:
I think it's a little confusing that the second question is called "Follow Up" in some instances, "Observations" in others. 

"Follow Up" language on the list page and the update page.
"Observations" elsewhere.

rebecca-caktus:
Good catch! Fixed now.

emullaney:
Looks great. The data spoofing form still uses "follow up" terminology as well. Should we update that, so that it's the same everywhere?

rebecca-caktus:
Updated again.

emullaney:
:green_apple:

PR #73, opened 2015-10-13
Raise error on form init for Org edit.

emullaney:
Test this out by creating an org with no API key, then going to Administration-> Organization, select "Edit" button. 

I'm not 100% sure on what to do within __init__ to raise an error. Let me know if there is a better way @rebecca-caktus.

### DIFF
--- a/tracpro/orgs_ext/middleware.py
+++ b/tracpro/orgs_ext/middleware.py
@@ -10,7 +10,7 @@ class HandleTembaAPIError(object):
     def process_exception(self, request, exception):
 
         rapidProConnectionErrorString = _(
-            "RapidPro appears to be down right now or we cannot connect due to your internet connection. "
+            "RapidPro appears to be down right now. "
             "Please try again later.")
 
         if isinstance(exception, TembaAPIError):
--- a/tracpro/orgs_ext/middleware.py
+++ b/tracpro/orgs_ext/middleware.py
@@ -1,16 +1,32 @@
 from django.utils.translation import ugettext_lazy as _
 from django.http import HttpResponseBadRequest
 
-from temba_client.base import TembaAPIError
+from temba_client.base import TembaAPIError, TembaConnectionError
 
 
 class HandleTembaAPIError(object):
     """ Catch all Temba exception errors """
 
     def process_exception(self, request, exception):
+
+        rapidProConnectionErrorString = _(
+            "RapidPro appears to be down right now or we cannot connect due to your internet connection. "
+            "Please try again later.")
+
         if isinstance(exception, TembaAPIError):
+            rapidpro_connection_error_codes = ["502", "503", "504"]
+
+            if any(code in exception.caused_by.message for code in rapidpro_connection_error_codes):
+                return HttpResponseBadRequest(
+                    rapidProConnectionErrorString)
+
+            else:
+                return HttpResponseBadRequest(
+                    _("Org does not have a valid API Key. "
+                      "Please edit the org through Site Manage or contact your administrator."))
+
+        elif isinstance(exception, TembaConnectionError):
             return HttpResponseBadRequest(
-                _("Org does not have a valid API Key. " +
-                  "Please edit the org through Site Manage or contact your administrator."))
+                rapidProConnectionErrorString)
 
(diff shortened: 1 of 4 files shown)


emullaney:
This is ready for re-review @rebecca-caktus

tracpro/orgs_ext/middleware.py
```diff
@@ -0,0 +1,16 @@
...
+from temba_client.base import TembaAPIError
+
+
+class HandleTembaAPIError(object):
+    """ Catch all Temba exception errors """
+
+    def process_exception(self, request, exception):
+        if isinstance(exception, TembaAPIError):
+            return HttpResponseBadRequest(
+                _("Org does not have a valid API Key. " +
```
rebecca-caktus: The plus sign between lines is not needed within parentheses.

tracpro/orgs_ext/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+from temba_client.base import TembaAPIError, TembaConnectionError
+
+
+class HandleTembaAPIError(object):
+    """ Catch all Temba exception errors """
+
+    def process_exception(self, request, exception):
+
+        rapidProConnectionErrorString = _(
+            "RapidPro appears to be down right now or we cannot connect due to your internet connection. "
```
rebecca-caktus: I don't think we need to include a remark about the client's internet connection since the server is the one reaching out to RapidPro.

emullaney: Riiiight. This is a really silly mistake!

tracpro/orgs_ext/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+            else:
+                return HttpResponseBadRequest(
+                    _("Org does not have a valid API Key. "
+                      "Please edit the org through Site Manage or contact your administrator."))
+
+        elif isinstance(exception, TembaConnectionError):
+            return HttpResponseBadRequest(
+                rapidProConnectionErrorString)
+
+        pass
```
rebecca-caktus: This could be `return None`.

tracpro/orgs_ext/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+    def process_exception(self, request, exception):
+
+        rapidProConnectionErrorString = _(
+            "RapidPro appears to be down right now. "
+            "Please try again later.")
+
+        if isinstance(exception, TembaAPIError):
+            rapidpro_connection_error_codes = ["502", "503", "504"]
+
+            if any(code in exception.caused_by.message for code in rapidpro_connection_error_codes):
```
rebecca-caktus: You can get the error code in numeric form like this.

```
if isinstance(exception, TembaAPIError):
    if isinstance(exception.caused_by, requests.HttpError):
        response = exception.caused_by.response
        if response:
            if response.status_code >= 500: 
                return HttpResponseBadRequest("message saying that's it's totally RapidPro's fault and you should try again later")
            elif response.status_code == 403:
                return HttpResponseBadRequest("message saying that you should fix your org's API key")

# Let it go. It's probably a problem with our code so a server error is sufficient.
return None                    
```

rebecca-caktus: On a side note with that, I propose that we only check for 5XX/TembaConnectionErrors and 403 errors. Everything else is probably specific to the view, either an error in our code :frowning: or something like a client data error that needs to be handled by validation in the view.

tracpro/orgs_ext/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+
+from temba_client.base import TembaAPIError, TembaConnectionError
+
+
+class HandleTembaAPIError(object):
+    """ Catch all Temba exception errors """
+
+    def process_exception(self, request, exception):
+
+        rapidProConnectionErrorString = _(
```
rebecca-caktus: By [convention](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables), Python variable names are written in lower case with underscores replacing spaces. :snake:

rebecca-caktus:
:rose: :fireworks: :gem:

PR #77, opened 2015-10-15
Remove nose requirement

rebecca-caktus:
There's no functional requirement that we use nose. If a developer would like to use it they can easily add it to their local settings file.


emullaney:
:+1:

PR #78, opened 2015-10-15
Contact Fields

rebecca-caktus:
* Administrator can use the org update page to choose which available fields (synced from RapidPro) should be visible on Tracpro.
* Available Fields are synced daily, as well as each time that a user loads the Org update page.
* Contact field values are synced to/from RapidPro.
* Chosen fields are visible on the Contact detail page.
* Chosen fields are editable on the Contact create/update pages.
* `Org.facility_code` has been removed; a migration adds this as an active data field on each org.

Modular commits to aid review. More tests forthcoming.

### DIFF
--- a/tracpro/contacts/models.py
+++ b/tracpro/contacts/models.py
@@ -8,7 +8,6 @@ from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 
-from dash.utils import intersection
 from dash.utils.sync import ChangeType
 
 from temba_client.types import Contact as TembaContact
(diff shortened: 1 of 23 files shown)


tracpro/contacts/models.py
```diff
@@ -127,53 +185,150 @@ def kwargs_from_temba(cls, org, temba_contact):
...
+        (TYPE_DATETIME, _("Datetime")),
+        (TYPE_STATE, _("State")),
+        (TYPE_DISTRICT, _("District")),
+    )
+
+    org = models.ForeignKey("orgs.Org")
+    label = models.CharField(max_length=255, blank=True)
+    key = models.CharField(max_length=255)
+    value_type = models.CharField(max_length=1, choices=TYPE_CHOICES)
+    show_on_tracpro = models.BooleanField(default=False)
```
emullaney: I like that all the fields exist, with a show_on_tracpro boolean. Good idea, rather than only tracking those that the user selects. I wouldn't have thought to do that.

emullaney:
Looks good, once test is fixed, I say :rocket:

PR #81, opened 2015-10-20
Fix to address distinct bug when sorting inbox messages on list page.

emullaney:


emullaney:
This is ready for review @rebecca-caktus. Thanks for the assistance working this out.

tracpro/msgs/views.py
```diff
@@ -112,7 +112,21 @@ class List(OrgPermsMixin, SmartListView):
...
-            qs = qs.order_by('contact', '-created_on').distinct('contact')
+
+            qs = qs.order_by('contact', '-created_on')
+            qs = qs.distinct('contact')
+
+            return qs
+
+        def order_queryset(self, qs):
+            """
+            Override the order_queryset() method from smartmin
```
rebecca-caktus: I don't think this comment is very helpful - maybe something more specific about how the default ordering mechanisms won't work due to the distinct query on contact?

Should we also consider applying validation of the `_order` parameter?

emullaney: Fixed!

rebecca-caktus:
:rose:

PR #82, opened 2015-10-21
Adding a delay to see if this fixes the issue of messages not showing up after send

emullaney:
@rebecca-caktus thoughts on this possible solution to https://trello.com/c/IAJAdNx4/66-feedback-inbox-messages-delay-in-appearing?

This will have to be tested on staging as locally this is not an issue.

### DIFF
--- a/tracpro/msgs/views.py
+++ b/tracpro/msgs/views.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, unicode_literals
 import logging
+import time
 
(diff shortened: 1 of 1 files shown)


rebecca-caktus:
Might help; won't hurt. Let's :shipit:

emullaney:
@rebecca-caktus one second seems to be enough of a delay to fix this. Can you test it as well, see if it works for you?

rebecca-caktus:
Looks good to me!

PR #83, opened 2015-10-23
BaselineTerm name modifications

rebecca-caktus:
Based on the discussion in our recent check-in, `BaselineTerms` should be called `Indicators` or `Recent Indicators` and "observation data" should be called "follow up data".


emullaney:
:apple:

PR #84, opened 2015-10-27
Dash updates for sync

emullaney:
1. Pointing to dash master repo. 
2. Updated sync_pull_contacts() method from dash with updated parameters.


rebecca-caktus:
:heavy_plus_sign:

PR #88, opened 2015-11-05
Fix unicode error

rebecca-caktus:
```
  File "/var/www/tracpro/source/tracpro/contacts/models.py", line 350, in set_value
    self.value = str(value)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2603' in position 0: ordinal not in range(128)
```

* Fixes the above unicode-related error when setting a `ContactField` value
* Change date parser to avoid `"1"` being parsed as `datetime.datetime(2015, 11, 1, 0, 0)`
* Adds tests for `ContactField`

### DIFF
--- a/tracpro/contacts/forms.py
+++ b/tracpro/contacts/forms.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import, unicode_literals
+
 from django import forms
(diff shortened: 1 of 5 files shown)


emullaney:
:sunny:

PR #89, opened 2015-11-06
Update package version requirements

rebecca-caktus:
* Update to latest Django security release
* Update all package versions. Notably, the latest Django did not seem to agree with django-celery==3.1.16
* Remove dependency on `django-reversion` since nothing required it and its usage had no effect


emullaney:
:grapes:

PR #91, opened 2015-11-11
Org languages

rebecca-caktus:
* User can use any language on any org.
* All languages are available in the lang switcher, but languages not in the org's `available_languages` are hidden by default.
* If the user has chosen a language through the lang switcher, it is always respected.
* If the user has not chosen a language, and is on an org page, then we activate the org's default language.
* If the user has not chosen a language, and is on a non-org page, then we allow Django to set the language (typically grabbed from the browser).
* Depends on [caktus/dash:org-languages](https://github.com/rapidpro/dash/compare/master...caktus:org-language).


locale/es/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,1083 @@
...
+
+#: tracpro/templates/polls/pollrun_participation.html:238
+msgid "Total"
+msgstr ""
+
+#: tracpro/templates/polls/response_by_pollrun.html:10
+#, python-format
+msgid ""
+"\n"
+"        These are the responses for <strong>%(region)s</strong>\n"
```
emullaney: I'm not super familiar with .po files... can you have variables in them?

emullaney:
LTR languages (Pashto) make the charts go a bit wonky. Not sure if that is part of this, or part of the charting fixes.

tracpro/templates/lang_switcher.html
```diff
@@ -1,18 +1,34 @@
...
+    {% if code == request.LANGUAGE_CODE %}
+      <li>{{ language }}</li>
+    {% else %}
+      <li{% if code not in show_languages %} class="hidden"{% endif %}>
         <form action="{% url "set_language" %}" method="POST">
           {% csrf_token %}
           <input type="hidden" name="next" value="{{ request.path }}" />
           <input type="hidden" name="language" value="{{ code }}" />
-          <a class="lang-submit" href="#" onclick="$(this).closest('form').submit();">
+          <button class="btn btn-link" type="submit">
```
emullaney: Good improvement here, switching from anchor to submit button. Simplified the code nicely.

emullaney:
Just a couple of questions really, seems ok to :ship: as long as we're okay with the wonky chart style on the homepage for LTR languages. If that's not ok, we'll want to adjust CSS for that before :rocket:.

locale/es/LC_MESSAGES/django.po
rebecca-caktus: Neither am I :smile_cat: As it turns out, yes you can:

https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#pluralization

rebecca-caktus:
I think that the wonkiness is a separate issue. Most orgs are currently using LTR languages so I think we can safely deal with the issue in the future.

PR #97, opened 2015-11-25
Show data in participants column

rebecca-caktus:
Data was not showing because the `chain` result could not be changed to a string.


emullaney:
:fireworks:

PR #102, opened 2015-11-30
Poll detail chart total

emullaney:
- Poll detail page has new drop down to display sum, average or response rate over date range
- Removed category charts for now, with plans to add them back in a more developer-friendly way
- This PR solves part of https://trello.com/c/hW01NlEt/70-poll-detail-page-for-numeric-questions-toggle-between-average-value-total-of-values-and-response-rate


tracpro/polls/charts.py
```diff
@@ -103,6 +105,63 @@ def multiple_pollruns(pollruns, question, regions):
...
+
+            complete_responses = count_by_status.get(True, 0)
+            total_responses = sum(count_by_status.values())
+            response_rates.append(round(100.0 * complete_responses / total_responses, 2))
+        else:
+            response_rates.append(0)
+    return response_rates
+
+
+def multiple_pollruns(pollruns, question, regions):
```
rebecca-caktus: This is not using the `regions` parameter (and probably should be).

rebecca-caktus:
:droplet: :closed_umbrella: :rainbow:

PR #104, opened 2015-12-03
Add profiles.user_self to ALLOW_NO_ORG

rebecca-caktus:
Possible mitigation for the potential of an infinite loop when a user requires a password change.


emullaney:
When I go directly to http://localhost:8000/ with no subdomain, I get redirected to http://localhost:8000/profile/self/ where I see this error:
Reverse for 'orgs.org_choose' with arguments '()' and keyword arguments '{}' not found. 0 pattern(s) tried: []

However, I do not see the error on subdomains.

rebecca-caktus:
You're right. I'd hoped this was a step in the right direction, but oh well. Closing.

PR #107, opened 2015-12-07
Average values should be rounded to 2 decimal places

emullaney:
Another small tweak. Average values should be rounded to 2 decimal places.

### DIFF
--- a/tracpro/polls/models.py
+++ b/tracpro/polls/models.py
@@ -635,7 +635,7 @@ class AnswerQuerySet(models.QuerySet):
                 # Append the sum total value for each date
                 if total:
                     answer_sums.append(total)
-                    answer_averages.append(round(float(total/total_answers), 2))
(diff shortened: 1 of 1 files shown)


tracpro/polls/models.py
```diff
@@ -635,7 +635,7 @@ def numeric_group_by_date(self):
                 # Append the sum total value for each date
                 if total:
                     answer_sums.append(total)
-                    answer_averages.append(float(total/total_answers))
+                    answer_averages.append(round(float(total/total_answers), 2))
```
rebecca-caktus: Change looks good. On an unrelated note the `float(...)` isn't doing anything since `total` is defined as a float above in line 630.

emullaney:
Noted. Removed the unnecessary float().

rebecca-caktus:
:star2:

PR #108, opened 2015-12-07
Display mean, standard deviation and response rate average on Poll Detail page

emullaney:
- Calculate mean and standard deviation via numpy across all average values in results
- Calculate response rate average via numpy across all responses rates in results


emullaney:
This is ready for review when you have time @rebecca-caktus

rebecca-caktus:
:ship: :anchor:

PR #109, opened 2015-12-09
Fixes issue with chart title not changing. 

emullaney:
Quick/small fix.
Issue was caused by multiple charts sharing the same ID. Fix in JS and ID value.


rebecca-caktus:
:+1:

PR #110, opened 2015-12-10
Editable Question type

rebecca-caktus:
* Track all Polls and Questions regardless of whether they are active, similar to DataField. (Note: we still only track responses for active polls.)
* Use formsets to edit `Question.question_type` and `Question.is_active` 
* Better tracking of custom name vs RapidPro name. Question/Poll names will update when the corresponding item on RapidPro is updated, unless a customization has been applied.

Before testing, make sure to run `sync_all_polls`.

### DIFF
--- a/tracpro/templates/polls/poll_update.html
+++ b/tracpro/templates/polls/poll_update.html
@@ -2,14 +2,29 @@
 
 {% load smartmin %}
 
-{% block post-fields %}
+{% block fields %}
+  <div class="form-group{% if form.name.errors %} error{% endif %}">
+    <label class="col-sm-2 control-label" for="{{ form.name.id_for_label }}">
+      {{ form.name.label }}
+    </label>
+    <div class="col-sm-10 smartmin-form-field">
+      {{ form.name|add_css:"form-control" }}
+    </div>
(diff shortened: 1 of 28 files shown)


tracpro/templates/polls/poll_update.html
```diff
@@ -0,0 +1,49 @@
+{% extends "smartmin/update.html" %}
+
+{% load smartmin %}
+
+{% block post-fields %}
+  {{ questions_formset.management_form }}
```
emullaney: I wonder if it would be helpful to include the rapidpro_name, not editable, but just informational, in case it's different from the local poll name?

tracpro/templates/polls/pollrun_read.html
```diff
@@ -11,10 +11,10 @@
 
   {% for question in questions %}
     <h3>
-      {{ question.order }}. {{ question.text }}
+      {{ question.order }}. {{ question.name }}
```
emullaney: I think you want {{ forloop.counter }} here

emullaney:
Minor adjustments and :shipit:

tracpro/templates/polls/poll_update.html
rebecca-caktus: Great idea!

PR #112, opened 2015-12-17
Question type updates

rebecca-caktus:
* Syncing a question with RapidPro should not overwrite the user-chosen question type.
* Guess the question type by inspecting the types of the rules in the associated ruleset.
* Mock the entire temba API client during tests.

The following code snippet can be run upon release to improve the accuracy of existing question types. I did not want to put this in a migration since it requires object methods which are not usually available in migrations.

```
from tracpro.polls.models import Poll
for poll in Poll.objects.all():
    for question in Question.objects.all():
        question.question_type = question._guess_question_type()
        question.save()
```

### DIFF
--- a/tracpro/contacts/tests/test_models.py
+++ b/tracpro/contacts/tests/test_models.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import, unicode_literals
 import datetime
 from decimal import Decimal
 
-import mock
 import pytz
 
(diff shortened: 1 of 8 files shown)


emullaney:
Just one suggestion - migration can be made with encapsulated code, or not, I don't feel strongly either way. 
:shipit:

PR #113, opened 2015-12-18
Require org subdomain

rebecca-caktus:
If an org's subdomain is the empty string, it is set as the request org when visiting a page on the global domain. (i.e., why we've been seeing "Somalia" on the top level of the staging site.)

This seems to be intentional behavior in Dash. But all of our orgs must be accessed through subdomains, so it seems logical to simply require that field when creating or editing an org.


emullaney:
:balloon:

PR #114, opened 2015-12-21
Improve Celery configuration for periodic tasks

rebecca-caktus:
We had several tasks that followed this general pattern:

```
@task
def do_something_for_all():
    for org in Org.objects.active():
        do_something_for_org(org.pk)  # synchronous

@task
def do_something_for_org(org_id):
    org = Org.objects.get(pk=org_id)
    # Do stuff for the org
    # Depending on the org size, this could take a few seconds to 30+ minutes
```

The `do_something_for_all` task was then added to the `CELERYBEAT_SCHEDULE`. In effect, each task in the `CELERYBEAT_SCHEDULE` could take 1-30 minutes and was scheduled every ~5 minutes. With only 2 workers on the deployment machines, we experienced unpredictable behavior.

Changes:

* Update `kombu` and `amqp` to resolve issues on Python 2.7.11 (`ImportError: cannot import name _uuid_generate_random`)
* `OrgTask`: Abstract task definition to use when when creating a task that operates on a single org.
* `ScheduleTaskForActiveOrgs`: Single task that schedules an `OrgTask` for each active org. This turns a 1-30 minute task into a collection of smaller, asynchronous tasks.
* Increased the number of workers running on deployment machines.


tracpro/settings/base.py
```diff
@@ -243,35 +243,48 @@
...
         'schedule': datetime.timedelta(minutes=5),
-        'args': (),
+        'kwargs': {
+            'task_name': 'tracpro.msgs.tasks.FetchOrgInboxMessages',
+        },
     }
 }
 
-CELERY_TIMEZONE = 'UTC'
+CELERYD_CONCURRENCY = 4
```
emullaney: Apparently this defaults to the number of CPU's on a machine. So would that be 1 normally? Just wondering.

rebecca-caktus: I think it's 2 for us on staging + production (both AWS smalls).

emullaney:
:+1:

tracpro/settings/base.py
ericnewcomer: This should likely be the number of cores, not discreet cpus. Even on your dev machine it's probably more than one. For example, our dev machines have 4.

PR #115, opened 2015-12-23
Delete polls that have been archived on RapidPro

rebecca-caktus:


emullaney:
Tested this out by creating a flow on rapidpro, selecting it on tracpro, then archiving it on rapidpro, then running sync_org_polls() which removed it successfully.

:shipit:

PR #116, opened 2015-12-30
Celery configuration

rebecca-caktus:
Done:

* All scheduler tasks are run every 5 minutes. (For consistency across tasks as we work out the configuration. We can adjust if need arises.)
* Scheduler tasks are handled on a separate queue, and expire after 5 minutes.
* Org tasks are handled on a separate queue, and expire after 5 minutes.
* Org tasks are locked and rate-limited to 5 minutes (per org).
* Turn off task prefetching (since all of our tasks make an API call and process results, thus are fairly long).
* Create another queue for org tasks, so that tasks kicked off in the code (like pollrun start and message sending) are not delayed.
* Created separate workers for each queue.


tracpro/orgs_ext/tasks.py
```diff
@@ -14,6 +19,12 @@
 
 logger = get_task_logger(__name__)
 
+LAST_RUN_KEY = 'last_run_time:{task}:{org}'
+
+ORG_TASK_LOCK = 'org_task:{task}:{org}'
+
+LOCK_EXPIRE = 60  # 1 minute
```
mlavin: This needs to be long enough for the task to finish (not sure if it is) otherwise it can be acquired without being released.

tracpro/orgs_ext/tasks.py
```diff
@@ -55,21 +66,56 @@ class OrgTask(PostTransactionTask):
...
+        return cache.delete(key)
+
+    def check_rate_limit(self, org):
+        """Return True if the task has been run for this org in the last 5 minutes."""
+        now = get_now()
+        last_run_key = LAST_RUN_KEY.format(task=self.__name__, org=org.pk)
+        last_run_time = cache.get(last_run_key)
+        if last_run_time is not None:
+            last_run_time = parse_iso8601(last_run_time)
+            if now - last_run_time < datetime.timedelta(minutes=5):
```
mlavin: Perhaps this offset should be a global like the key formats and timeout.

rebecca-caktus:
Updated. Thanks @mlavin.

mlavin:
:pineapple:

PR #117, opened 2015-12-31
Categorical Charts and Word Clouds on Poll Detail Page

emullaney:
Non numeric charts on poll detail page
- Categorical flat area graphs for multiple choice questions (as before)
- Word clouds for free text questions
- Template is now smart enough to detect question type and display the correct chart type (numeric, categorical or word cloud)
- Angular js is no longer used for charting. Will remove in future pull request.
- Increased data window to 150 days (this did not come from a requirement, but I needed to see older data during my testing and thought it was useful)
- New charting should be much easier to understand for future code improvements/forks


tracpro/templates/frame.html
```diff
@@ -29,6 +29,7 @@
       <![endif]-->
 
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js"></script>
+    <script type="text/javascript" src="{{ STATIC_URL }}js/jqcloud-1.0.4.min.js"></script>
```
rebecca-caktus: Why did this need to move?

emullaney: I'm not sure why exactly but for some reason the word cloud would not work until I moved it.

tracpro/templates/polls/poll_read.html
```diff
@@ -45,10 +45,14 @@
   {% for question in questions %}
     <h3>
       {{ forloop.counter }}. {{ question.name }}:
-      <span id="question-{{ question.pk }}-heading-value-type">{{ value_type_selected|title }}</span>
+      {% if question.question_type == question.TYPE_NUMERIC %}
+        <span id="question-{{ question.pk }}-heading-value-type">{{ value_type_selected|title }}</span>
+      {% elif question.question_type == question.TYPE_MULTIPLE_CHOICE %}
+        <span>Categorical</span>
+      {% endif %}
```
rebecca-caktus: Having no title looks fine.  However, is there something that would make sense for open-ended questions?

tracpro/templates/polls/poll_read.html
```diff
@@ -109,10 +113,73 @@
...
+              ]
+          });
+        });
+      </script>
+
+    {% elif question.question_type == question.TYPE_OPEN and question.pollrun_dict|length > 2 %}
+      <!-- word cloud chart -->
+      <div id='chart_{{ question.pk }}' class='chart' ></div>
+      <script type="text/javascript">
+        var words = {{ question.pollrun_dict|safe }};
```
rebecca-caktus: This is executing while the page loads - possibly the reason that you needed to move `jcloud` to the top of the page?

rebecca-caktus:
Awesome :smile_cat:

tracpro/templates/polls/poll_read.html
emullaney: I think I will add a title to the open ended clouds...

tracpro/templates/polls/poll_read.html
emullaney: Yeah, that makes sense. I guess this wasn't an issue when it was in the coffee script.

PR #119, opened 2016-01-04
New tests for numeric charts. Fix up non numeric chart tests.

emullaney:
Adding numeric chart tests to the test suite. Going to merge this into the other open pull request #117  immediately for review as part of that request.


emullaney:
Merged in directly so it can be reviewed as part of #117

PR #120, opened 2016-01-04
Back button bug fix for chart data

emullaney:
Issue described in https://basecamp.com/1849796/projects/9585844/messages/52475672, copied below:

"one thing i noticed is that the filters are a bit funky if you change from sum to response rate, click and poll date, and then use the back button to return to the poll summary (e.g. https://zimbabwe.edutrac-staging.cakt.us/poll/read/42/). it will display sum, but the filter still shows response rate"

Update: 
This fix calls the onValueTypeChange() function on initial page load so that when the user clicks the Back button, the correct charts are loaded.

### DIFF
--- a/tracpro/templates/polls/poll_read.html
+++ b/tracpro/templates/polls/poll_read.html
@@ -142,8 +142,13 @@
           $("#question-{{ question.pk }}-heading-value-type").html("Response Rate");
         {% endfor %}
       }
-    }).change()
-    // Run this once on initial page load via .change(),
+    })
+
+    // Run this once on initial page load,
     // in the case that the user has hit the Back button.
+    $(function () {
+      $('#value-type-select').change();
+    });
+
 </script>
 {% endblock %}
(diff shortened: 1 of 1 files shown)


tracpro/templates/polls/poll_read.html
```diff
@@ -114,37 +115,44 @@
...
+    });
     function onWindowChange(ctrl) {
       ctrl.form.submit();
     }
-    function onValueTypeChange(ctrl) {
+    function onValueTypeChange() {
       valueType = $("#value-type-select :selected").text().toLowerCase();
+      changeChartValueType(valueType);
+      // Keep track of this value in a "cached" hidden variable for Back button click
+      $("#value-type-cache").val(valueType);
```
rebecca-caktus: I see that you are setting the value of the cache here, but I don't see where it is used.

emullaney: Huh. You are right. I was using it, but then I guess I didn't need it after all. Checking into this...

emullaney: It was unnecessary and has been removed!

rebecca-caktus:
I find that this is a neat way to attach an event function and trigger it to run on page load:

```
$('#value-type-select').on('change', function() {
  var val = $(this).val().lower();
  // ...
}).change()
```

rebecca-caktus:
Good change :strawberry:

PR #121, opened 2016-01-05
PollRun Participation tab display participating groups/regions only

emullaney:
Removing line that adds non-participating groups/regions to the participation dictionary.

Fixes https://trello.com/c/iZqoS4FI/92-check-to-see-if-participation-tab-on-pollrun-detail-page-is-displaying-groups-that-were-not-sent-poll-do-not-display-these-group


rebecca-caktus:
:chocolate_bar:

PR #123, opened 2016-01-05
Filter by arbitrary date range

rebecca-caktus:
* Use form to manage poll chart options
* Filter by custom date range
* Decouple javascript from poll read template

In the future, we might consider:
* look into using a bootstrap form rendering library
* better form interface
* JSON endpoint for chart data


emullaney:
:rabbit: :tophat:

PR #124, opened 2016-01-05
small fix for chart js

emullaney:
Small adjustment to deal with setting series data only on numeric charts via the drop down


rebecca-caktus:
:+1:

PR #126, opened 2016-01-08
Poll chart filters: Add pre-defined date ranges for 6 and 12 months

rebecca-caktus:


tracpro/polls/forms.py
```diff
@@ -118,8 +120,9 @@ def clean(self):
                 if window == 'month':
                     start_date, end_date = get_month_range()
                 else:
+                    number, unit = window.split('-')  # e.g., 6-months
```
emullaney: clever use of split() with relativedelta().

emullaney:
:watch:

PR #127, opened 2016-01-08
Poll chart filters: Configure datepicker options, including drop-down year selection

rebecca-caktus:
These are the options we have available: https://api.jqueryui.com/datepicker/#option-yearSuffix

### DIFF
--- a/tracpro/static/js/poll-charts.js
+++ b/tracpro/static/js/poll-charts.js
@@ -96,7 +96,7 @@ $(function() {
         changeYear: true,
         selectOtherMonths: true,
         showOtherMonths: true,
-        yearRange: "2010:" + new Date().getFullYear()
+        yearRange: "2013:" + new Date().getFullYear()
     });
 
     /* Update button text when filter form display is toggled. */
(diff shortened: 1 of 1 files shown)


tracpro/static/js/poll-charts.js
```diff
@@ -91,7 +91,13 @@ jQuery.fn.extend({
 
 $(function() {
     /* Initialize date fields. */
-    $("input[class^='datepicker']").datepicker();
+    $("input[class^='datepicker']").datepicker({
+        changeMonth: true,
+        changeYear: true,
+        selectOtherMonths: true,
+        showOtherMonths: true,
+        yearRange: "2010:" + new Date().getFullYear()
```
emullaney: Any reason why we started with 2010? I'm not sure RapidPro has been around that long, looks like TextIt was introduced in 2013... so we'd most likely have no data before 2013.

emullaney:
A question then :ship:

tracpro/static/js/poll-charts.js
rebecca-caktus: Good thinking. 2013 seems like a great bet.

PR #128, opened 2016-01-08
Poll chart filters: Allow user to filter by only start date or only end date

rebecca-caktus:


tracpro/polls/forms.py
```diff
@@ -99,9 +99,10 @@ def clean(self):
             # User must provide specific start or end dates.
             start_date = self.cleaned_data.get('start_date')
             end_date = self.cleaned_data.get('end_date')
-            if not (start_date and end_date):
-                self.add_error('start_date', _("Please select a start date."))
-                self.add_error('end_date', _("Please select an end date."))
+            if not (start_date or end_date):
+                self.add_error(
+                    forms.ALL_FIELDS,
+                    _("Please choose a start date or an end date."))
```
emullaney: If I choose a start date and blank out the end date, I get an error.

rebecca-caktus:
Good catch; updated.

emullaney:
:calendar:

PR #132, opened 2016-01-12
Filter refactoring

rebecca-caktus:
Some refactors ahead of upcoming filtering changes. Notably:

* Reduce filter form complexity by always binding form
* Change `data_type` -> `num_display` because I want to use the `data_` prefix to search on DataFields as required by the upcoming ticket
* Created common `get_answers()` method for the charting code so that we have one place to apply filtering changes
* Ensure database query counts are not dependent upon the number of pollruns (closer to a constant # of queries)


emullaney:
Great improvements overall, once develop is merged back in :shipit:

PR #134, opened 2016-01-13
Filter by contact data field

rebecca-caktus:
Modular commits should aid review. Main goal of this was to factor out answer filters and add filtering by contact data field value. 

One other thing of note is that I think that answer sums/averages should be created by pollrun date rather than response__created_on (0ceddbc). For the sake of time this week I've kept the change in this pull request. It would be possible to factor out that change if needed.


emullaney:
looks great! :boat:

PR #135, opened 2016-01-13
Fix incorrect super() reference

rebecca-caktus:
Discovered via this log message:

```
Jan 13 12:33:58 tracpro.staging celery-beat.log:  SchedulingError: Couldn't apply scheduled task sync-contacts: super(type, obj): obj must be an instance or subtype of type
```


emullaney:
:palm_tree:

PR #136, opened 2016-01-13
Filters UI adjustments

rebecca-caktus:
Some simple changes that I hope make the filters form friendlier to use. Still very open to suggestions on making it better.


emullaney:
:cinema:

PR #137, opened 2016-01-13
Break filter form into reusable bits

rebecca-caktus:
Prerequisite work toward adding filters to baseline terms.


rebecca-caktus:
This could use some more work; closing for now.

PR #138, opened 2016-01-13
Handle baseline options + filtering using forms

rebecca-caktus:
Based off of the work in #137. More prerequisite work for adding chart filters to baseline terms.

Code diff until #137 is merged: https://github.com/rapidpro/tracpro/compare/reusable-filters...baseline-filter-form


rebecca-caktus:
This could use some more work; closing for now.

PR #139, opened 2016-01-19
Pollrun Charts

emullaney:
Bar charts on the single pollrun detail page. 
Word clouds for open results. 
Unit tests in progress (adding shortly).


daaray:
* Tests pass
 * Smoke test passes for verifying different chart types
 * Insofar as I understand the code, it looks great

:100:

PR #140, opened 2016-01-19
PollRun mean standard deviation and response rate

emullaney:
Calculate mean, standard deviation and response rate on each pollrun
Display these values
- refactor this display into an include that gets re-used on poll detail page


daaray:
:100:

PR #142, opened 2016-01-20
Sort categories by numeric values or strings

emullaney:
Poll Detail - all multiple choice questions
Pollrun Detail - all results that aren't open-ended

### DIFF
--- a/tracpro/polls/charts.py
+++ b/tracpro/polls/charts.py
@@ -168,6 +168,7 @@ def multiple_pollruns_multiple_choice(answers, pollruns, question):
         })
 
     dates = [pollrun.conducted_on.strftime('%Y-%m-%d') for pollrun in pollruns]
+
     return {
         'dates': dates,
         'series': series,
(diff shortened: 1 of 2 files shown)


daaray:
Validated that categories are now sorted:

sorted_categories

Is it possible to add a quick test for this change as well?

rebecca-caktus:
Great result!

`sort_by_category` is used after `category_counts_by_pollrun`, so maybe the category sorting could be completed in that function?

rebecca-caktus:
You might also find the `key` argument to `sort()` relevant here:

```
def natural_category_key(category):
    # return what should actually be sorted on

list_of_things.sort(key=natural_category_key)
list_of_tuples.sort(key=lambda (a, b): natural_category_key(a))
```

emullaney:
I've added a test per @daaray's suggestion.

@rebecca-caktus's  refactoring suggestions are excellent, however, I'd like to push this up now, and look at those suggestions at a later date, so we can get this fix out.

daaray:
:fireworks: :fireworks: :fireworks:

PR #144, opened 2016-01-21
Fixes #141

### LINKED ISSUE #141: Dev Setup Docs
There are a few notes which should be added to the dev setup docs: http://tracpro.readthedocs.org/en/latest/developers/local-developer-setup.html

- As part of step 3 (before running migrate is step 4) you need to deactivate and reactivate the virtualenv to pick up the environment variable for the settings ` deactivate && workon tracpro`
- You need to have `coffeescript` and `lessc` installed for the precompilers `sudo npm install -g coffee-script less`
- When creating a superuser locally (not explicitly in the docs) you need to ensure the password meets the requirements or else you won't be allowed to log in

daaray:
add notes to dev setup regarding superuser, required node packages, and when to deactivate/reactive the venv

### DIFF
--- a/docs/source/developers/local-developer-setup.rst
+++ b/docs/source/developers/local-developer-setup.rst
@@ -77,7 +77,7 @@ may be needed for Mac setup. To begin you should have the following applications
 
 #. Create Super User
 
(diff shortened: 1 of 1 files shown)


emullaney:
:coffee:

emullaney:
:+1:

PR #146, opened 2016-01-22
HT-15 Breakout Deployment

mlavin:
Removing deployment related pieces from this repo which will be managed separately going forward. Remaining work to update/move docs as well.


emullaney:
There are references to the "conf" folder in https://github.com/rapidpro/tracpro/blob/develop/Makefile, should those be updated or moved to hosted-tracpro?

mlavin:
Yes those should be removed here and moved as well. Good catch.

mlavin:
Updated.

emullaney:
:recycle:

PR #147, opened 2016-01-22
Reusable filters

rebecca-caktus:
Redux of #137. Preliminary work toward reusing filters for baseline terms.


emullaney:
:coffee:

PR #149, opened 2016-01-25
Category sort refactor

rebecca-caktus:


tracpro/polls/utils.py
```diff
@@ -49,3 +49,9 @@ def extract_words(text, language):
     words = re.split(r"[^\w'-]", text.lower(), flags=re.UNICODE)
     ignore_words = ignore_words
     return [w for w in words if w not in ignore_words and len(w) > 1]
+
+
+def natural_sort_key(text):
+    """Key to sort text in a humanized way, e.g., 11 should come before 100."""
+    return [int(t) if t.isdigit() else t.lower()
+            for t in re.split('([0-9]+)', text)]
```
emullaney: I find this a bit hard to read.. I preferred the old way! Does this work faster/better?

rebecca-caktus: This version catches more edge cases, for example:

* When a letter comes before the number: `["b1", "a1"]`
* When there are multiple numbers in the text: `["10-100", "10-11"]`

This method uses turns the text input into a list of parts, where each numeric part is cast as an integer: `"a11b12"` becomes `["a", 11, "b", 12]`. This can be used as a sort key, taking advantage of [how Python compares lists](http://www.tutorialspoint.com/python/list_cmp.htm).

emullaney: I was pretty sure the previous version took care of the ["10-100", "10-11"] example, but maybe it didn't. That's fine. I still find this somewhat unreadable. Any reason not to split it out into multiple lines for clarity with a separate return statement?

emullaney:
once travis is done.... :rocket:

PR #150, opened 2016-01-25
Adding mean, stdev and response rate calculations

emullaney:
Added calculations to  to poll and poll detail for non numeric questions (everything except open-ended).
There is one flake8 issue I'm not sure how to address.
https://trello.com/c/oBDHuLAN/102-add-mean-stdev-response-rate-to-categorical-questions-in-addition-to-numeric-questions

### DIFF
--- a/tracpro/polls/charts.py
+++ b/tracpro/polls/charts.py
@@ -53,7 +53,7 @@ def single_pollrun(pollrun, question, answer_filters):
 
     # Calculate the average, standard deviation,
     # and response rate for this pollrun
-    if question.question_type == Question.TYPE_NUMERIC:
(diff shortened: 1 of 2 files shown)


rebecca-caktus:
You don't need to save the results to a variable if you are only looking for the side effect.

rebecca-caktus:
I know what this code is supposed to do, but I think newcomers would find it confusing. (And I might too, in a month or two.) A lot of unrelated calculation is being done to get the side effect. Do you have any ideas for how we could factor out the useful bit?

rebecca-caktus:
Thanks for the comment!

As discussed with @emullaney, refactoring will probably be addressed in upcoming work for baseline term filters. For now, :turtle:

PR #154, opened 2016-02-02
Chart toggle regions

emullaney:
Split out regions for poll charts, numeric only. 
Completes https://trello.com/c/M3DZEXVt/81-the-ability-to-toggle-between-aggregated-data-and-disaggregated-line-graphs-by-sub-region
Looking into adding tests, but this is ready to check over @rebecca-caktus


emullaney:
@rebecca-caktus please review when you have a moment. Thanks!

tracpro/polls/models.py
```diff
@@ -772,6 +792,24 @@ def category_counts_by_pollrun(self):
...
+    def get_answer_summaries_regions(self):
+        """Return the sum and average of numeric answers to each pollrun."""
+        """Split out values by regions"""
+        answer_sums_list = []
+        answer_avgs_list = []
+
+        for region in self.get_regions():
+            region_answers = self.filter(response__contact__region=region)
+            # Get the dictionaries of sums and averages per pollrun
+            answer_sums, answer_avgs = region_answers.get_answer_summaries()
```
rebecca-caktus: This loop is generating queries on each iteration, resulting a 5-6x slower page load. I think that doing the iteration in Python would save a lot of that time.

tracpro/polls/charts.py
```diff
@@ -103,22 +107,52 @@ def multiple_pollruns_multiple_choice(pollruns, answers):
     }
 
 
-def multiple_pollruns_numeric(pollruns, responses, answers, question):
+def multiple_pollruns_numeric(pollruns, responses, answers, question, split_regions):
     """Chart data for multiple pollruns of a poll."""
-    answer_sums, answer_avgs = answers.get_answer_summaries()
-    response_rates = responses.get_response_rates()
+    if split_regions:
```
rebecca-caktus: This function acts in two very different ways based on the value of `split_regions`. I think it would be an easier read if split into two functions.

tracpro/static/js/poll-charts.js
```diff
@@ -39,6 +39,62 @@ jQuery.fn.extend({
...
+                        name: data['region-list'][i],
+                        data: data[dataType][i]
+                    });
+                }
+                chart.highcharts({
+                    chart: {
+                        type: 'area'
+                    },
+                    title: {
+                        text: 'Stacked area chart'
```
rebecca-caktus: This isn't a very specific title, and none of the other charts have `title` values passed in.

tracpro/polls/models.py
emullaney: Any ideas on how to mitigate this off the top of your head? I'm not sure how to refactor this.

tracpro/static/js/poll-charts.js
emullaney: Fixed.

tracpro/polls/charts.py
emullaney: Split this into 2 functions.

emullaney:
Refactored and ready for review again @rebecca-caktus

rebecca-caktus:
:shipit:

PR #155, opened 2016-02-03
HT-21 Updating to latest and greatest caktus project settings with new .env set up

emullaney:


docs/source/developers/local-developer-setup.rst
```diff
@@ -46,9 +46,7 @@ may be needed for Mac setup. To begin you should have the following applications
 #. Create a local settings file and set your DJANGO_SETTINGS_MODULE to use it and restart the environment so that the change will take effect::
```
rebecca-caktus: Text here should be updated now that the environment no longer needs to be restarted.

emullaney: Feel free to update it @rebecca-caktus - I'm not entirely sure what it should say.

rebecca-caktus:
:+1:

PR #156, opened 2016-02-09
Fix for timedelta/isoformat bug. Update celery to latest.

emullaney:


tracpro/settings/dev.py
```diff
@@ -3,7 +3,7 @@
 from .base import *  # noqa
 
 
-CELERY_ALWAYS_EAGER = True
+CELERY_ALWAYS_EAGER = False
```
rebecca-caktus: Let's not commit this change. Although it masked this particular error, using `CELERY_ALWAYS_EAGER = True` is usually a benefit for development.

emullaney: :+1:

rebecca-caktus:
:+1:

PR #157, opened 2016-02-12
Maps

emullaney:
Code for API sync borrowed heavily from https://github.com/rapidpro/ureport/blob/master/ureport/locations/models.py

This might not be quite ready, but I wanted to open it for now.


rebecca-caktus:
It looks like we duplicated efforts for syncing :frowning:

PR #158, opened 2016-02-17
Display data from active regions only 

emullaney:
-poll chart
-pollrun charts
-pollrun participation tab
-pollrun responses tab

### DIFF
--- a/tracpro/polls/models.py
+++ b/tracpro/polls/models.py
@@ -542,7 +542,7 @@ class PollRun(models.Model):
         if not self.covers_region(region, include_subregions):
             raise ValueError(
                 "Request for responses in region where poll wasn't conducted")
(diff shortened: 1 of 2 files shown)


rebecca-caktus:
:+1: from me.

For posterity: I looked into whether we need to update `request.data_regions` in  `UserRegionsMiddleware` to exclude inactive regions. This is how it gets set:

```
request.data_regions = request.region.get_descendants(include_self=True)
request.data_regions = request.data_regions.filter(pk__in=request.user_regions)
```

`request.user_regions` ultimately uses this query:

```
regions = user.get_direct_regions(org).get_descendants(include_self=True)
regions = regions.filter(is_active=True)
```

So it looks like `request.data_regions` is already doing the right thing.

PR #159, opened 2016-02-22
Baseline filters

rebecca-caktus:


emullaney:
Bunch of happy comments, one thing to fix (filtering out inactive contacts) then :anchor:

PR #162, opened 2016-02-29
Update 404 template

rebecca-caktus:
* Extend from base.
* Remove reference to non-existent image.

To test locally, set:

```
DEBUG = False
ALLOWED_HOSTS = ['localhost', 'org1.localhost', 'org2.localhost']
```

and run the server with `python manage.py runserver --insecure`.


emullaney:
:ship:

PR #164, opened 2016-03-01
Only retrieve 7 days of non-archived inbox messages

emullaney:
In order to speed up this task


rebecca-caktus:
:mailbox_with_no_mail:

PR #166, opened 2016-03-11
Email config

rebecca-caktus:
Grab email configuration values from the environment.

Also included:
* Utilities for environment variables
* Use Django defaults if possible
* Remove need for `HOSTNAME` environment variable (since it is a duplicate of `DOMAIN`)


mlavin:
There are plenty of other settings that should be made configurable in a similar fashion (particularly the cache/broker settings) but one thing at a time. This is a nice pattern. :cake:

PR #168, opened 2016-03-11
Settings tweaks

rebecca-caktus:
* `SITE_HOST_PATTERN` appears to be a historic setting, no longer in use in this repo, `rapidpro`, `rapidpro-python`, `dash`, or `smartmin`.
* Several other settings are based on the domain and can thus be generated.
* Require specification of `DB_NAME` and `DB_USER` from the environment.


emullaney:
:100: 
Fyi - I just had to be on the email-config branch of hosted-tracpro

PR #169, opened 2016-03-11
Update logging configuration

rebecca-caktus:
Tweaks to the logging configuration based on our current hosted Tracpro setup.

The console is written to syslog, which we then forward to a log service.


mlavin:
At some point it might be easier to just configure the `root` handler rather than all of the individual handlers. Doesn't need to block this improvement :sunglasses:

rebecca-caktus:
@mlavin I added 0a7f0cd - is this what you mean?

mlavin:
That was it :smile:

PR #170, opened 2016-03-15
Use unicode_literals in all files

rebecca-caktus:
Built on the work of @emullaney in 3ea56c6 and added `from __future__ import unicode_literals` to all Python files that don't already have it.


emullaney:
:watermelon: :sunny: :shell:

PR #171, opened 2016-03-18
Considerations when designing flows for TracPro

rebecca-caktus:
```
pip install sphinx
cd docs/
make html
xdg-open build/html/users/designing-flows.html
```

The final command is a shortcut to open the generated HTML file in the browser.


emullaney:
well written
:books:

PR #176, opened 2016-03-28
Task backoff

rebecca-caktus:
If a task fails, exponentially increase the timeout before it can run again.

### DIFF
--- a/tracpro/orgs_ext/tasks.py
+++ b/tracpro/orgs_ext/tasks.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import datetime
+import logging
 
 from celery import signature
 from celery.exceptions import SoftTimeLimitExceeded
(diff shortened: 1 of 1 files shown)


tracpro/orgs_ext/tasks.py
```diff
@@ -140,6 +163,9 @@ def run(self, org_pk):
                 msg = "{}: Time limit exceeded for {}".format(self.__name__, org.name)
                 logger.error(msg)
 
+                failure_count_key = self.get_cache_key(org, FAILURE_COUNT)
+                cache.set(failure_count_key, 1 + cache.get(failure_count_key, 0))
```
mlavin: `cache.incr` isn't atomic on all backends but should be for memcache/redis and would preferable to use here.

tracpro/orgs_ext/tasks.py
```diff
@@ -88,19 +94,30 @@ def apply_async(self, *args, **kwargs):
...
             last_run_time = parse_iso8601(last_run_time)
-            if now - last_run_time < settings.ORG_TASK_TIMEOUT:
-                return True
+            delta = settings.ORG_TASK_TIMEOUT * 2 ** cache.get(fail_count_key, 0)
+            delta = max(delta, datetime.timedelta(days=1))
+            if now - last_run_time < delta:
+                return last_run_time + delta
+
+        # Set the current time as the last run time and allow the task to run now.
         cache.set(last_run_key, format_iso8601(now))
```
mlavin: This appears to use the default cache timeout (I don't see it configured so it would be 5 mins by default) meaning this key will expire before it should be released.

rebecca-caktus: @mlavin We set `expires` here; is this sufficient? https://github.com/rapidpro/tracpro/pull/176/files#diff-3ee0e4667e5c0a8507a252abb50e8ce5R89

mlavin: That is the expiry for the task message but not for this cache key. https://docs.djangoproject.com/en/1.9/ref/settings/#timeout

rebecca-caktus: Thanks for the clarification! This PR is updated.

mlavin:
:rocket:

PR #178, opened 2016-03-28
RapidPro uuid is only unique together with another item

rebecca-caktus:
Three locations have unique constraints that are better represented as `unique_together`.


emullaney:
:shaved_ice:

PR #185, opened 2016-05-02
Don't set ADMINS when deploying to Vagrant

rebecca-caktus:
To stop the :umbrella: of emails during Vagrant setup :smile_cat:


daaray:
👍  ha; sorry!

PR #186, opened 2016-05-02
Check environment variables for custom staticfiles & template dirs

daaray:
This change allows a provisioning/deployment wrapper to utilize custom templates and static files.

### DIFF
--- a/tracpro/settings/deploy.py
+++ b/tracpro/settings/deploy.py
@@ -73,10 +73,10 @@ SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
 STATIC_ROOT = os.path.join(WEBSERVER_ROOT, 'public', 'static')
 
 if from_env('CUSTOM_STATICFILES_DIR'):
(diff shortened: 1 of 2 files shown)


tracpro/settings/deploy.py
```diff
@@ -72,6 +72,12 @@
 
 STATIC_ROOT = os.path.join(WEBSERVER_ROOT, 'public', 'static')
 
+if from_env('CUSTOM_STATICFILES_DIR'):
+    STATICFILES_DIRS.insert(0, os.path.join(WEBSERVER_ROOT, from_env('CUSTOM_STATICFILES')))
```
mlavin: I think the environment variables should contain the full path rather than a directory relative to `WEBSERVER_ROOT` which isn't configurable.

daaray: Thanks for the feedback; I will make those changes.

emullaney:
:shaved_ice: :sunglasses:

PR #191, opened 2017-01-17
Fix timeout on poll selection

emullaney:
Addresses https://caktus.atlassian.net/browse/TD-8

This is a proof of concept fix that I'd like to test out on staging.

Because the Poll sync also pulls down all question/categorical data down locally, it can time out when there are many polls for an org with a lot of questions. 

This change moves the question sync into the form save so that only active polls get questions synced.

### DIFF
--- a/tracpro/polls/forms.py
+++ b/tracpro/polls/forms.py
@@ -54,20 +54,10 @@ class ActivePollsForm(forms.Form):
         uuids = self.cleaned_data['polls'].values_list('flow_uuid', flat=True)
         models.Poll.objects.set_active_for_org(self.org, uuids)
 
-        # Save the associated Questions for this poll here
-        # now that these polls have been activated for the Org
-        selected_poll_names, selected_polls = [], []
-        for poll in self.cleaned_data['polls']:
-            selected_poll_names.append(poll.name)
-            selected_polls.append(models.Poll.objects.get(id=poll.id))
-
-        temba_polls = self.org.get_temba_client().get_flows(archived=False)
-        temba_polls = {p.uuid: p for p in temba_polls}
-
         # Call a celery task to update the questions and categories
         # This takes a long time, so let's schedule it to run in the background
         sync_questions_categories.delay(
-            temba_polls, selected_poll_names, selected_polls)
+            self.org, self.cleaned_data['polls'])
 
 
 class PollChartFilterForm(filters.DateRangeFilter, filters.DataFieldFilter,
--- a/tracpro/polls/forms.py
+++ b/tracpro/polls/forms.py
@@ -8,6 +8,7 @@ from dash.utils import get_month_range
 from tracpro.charts import filters
 
 from . import models
+from .tasks import sync_questions_categories
 
 
 class PollForm(forms.ModelForm):
@@ -52,6 +53,21 @@ class ActivePollsForm(forms.Form):
     def save(self):
         uuids = self.cleaned_data['polls'].values_list('flow_uuid', flat=True)
         models.Poll.objects.set_active_for_org(self.org, uuids)
+
+        # Save the associated Questions for this poll here
+        # now that these polls have been activated for the Org
+        selected_poll_names, selected_polls = [], []
+        for poll in self.cleaned_data['polls']:
+            selected_poll_names.append(poll.name)
+            selected_polls.append(models.Poll.objects.get(id=poll.id))
+
+        temba_polls = self.org.get_temba_client().get_flows(archived=False)
+        temba_polls = {p.uuid: p for p in temba_polls}
+
+        # Call a celery task to update the questions and categories
+        # This takes a long time, so let's schedule it to run in the background
+        sync_questions_categories.delay(
+            temba_polls, selected_poll_names, selected_polls)
 
 
 class PollChartFilterForm(filters.DateRangeFilter, filters.DataFieldFilter,
--- a/tracpro/polls/tasks.py
+++ b/tracpro/polls/tasks.py
@@ -140,22 +140,18 @@ def sync_questions_categories(org, polls):
 
     # Save the associated Questions for this poll here
     # now that these polls have been activated for the Org
-    selected_poll_names, selected_polls = [], []
-    for poll in polls:
-        selected_poll_names.append(poll.name)
-        selected_polls.append(Poll.objects.get(id=poll.id))
+    selected_poll_names = [poll.name for poll in polls]
 
     temba_polls = org.get_temba_client().get_flows(archived=False)
     temba_polls = {p.uuid: p for p in temba_polls}
 
-    total_polls = len(selected_poll_names)
+    total_polls = len(polls)
     logger.info(
         "Retrieving Questions and Categories for %d Poll(s) that were recently updated via the interface." %
         (total_polls))
     for temba_poll in temba_polls.values():
         if temba_poll.name in selected_poll_names:
-            poll_index = selected_poll_names.index(temba_poll.name)
-            poll = selected_polls[poll_index]
+            poll = polls[selected_poll_names.index(temba_poll.name)]
             # Sync related Questions, and maintain question order.
             temba_questions = OrderedDict((r.uuid, r) for r in temba_poll.rulesets)
 
--- a/tracpro/polls/tasks.py
+++ b/tracpro/polls/tasks.py
@@ -134,13 +134,23 @@ class SyncOrgPolls(OrgTask):
 
 
 @task
-def sync_questions_categories(temba_polls, selected_poll_names, selected_polls):
+def sync_questions_categories(org, polls):
     # Create new or update SELECTED Polls to match RapidPro data.
-    from tracpro.polls.models import Question
+    from tracpro.polls.models import Poll, Question
+
+    # Save the associated Questions for this poll here
+    # now that these polls have been activated for the Org
+    selected_poll_names, selected_polls = [], []
+    for poll in polls:
+        selected_poll_names.append(poll.name)
+        selected_polls.append(Poll.objects.get(id=poll.id))
+
+    temba_polls = org.get_temba_client().get_flows(archived=False)
+    temba_polls = {p.uuid: p for p in temba_polls}
 
     total_polls = len(selected_poll_names)
     logger.info(
-        "Retrieving Questions and Categories for %d Polls that were recently updated via the interface." %
+        "Retrieving Questions and Categories for %d Poll(s) that were recently updated via the interface." %
         (total_polls))
     for temba_poll in temba_polls.values():
         if temba_poll.name in selected_poll_names:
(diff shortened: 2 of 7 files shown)


tracpro/polls/views.py
```diff
@@ -191,6 +191,31 @@ def get_form_kwargs(self):
...
+
+            org = self.request.org
+            temba_polls = org.get_temba_client().get_flows(archived=False)
+            temba_polls = {p.uuid: p for p in temba_polls}
+
+            # Create new or update SELECTED Polls to match RapidPro data.
+            for temba_poll in temba_polls.values():
+                if temba_poll.name in selected_polls:
+                    # Sync related Questions, and maintain question order.
+                    temba_questions = temba_poll.rulesets
```
dpoirier: This line seems unnecessary, since the next line assigns a different value to the same variable.

tracpro/polls/views.py
```diff
@@ -191,6 +191,31 @@ def get_form_kwargs(self):
...
+                    temba_questions = OrderedDict((r.uuid, r) for r in temba_poll.rulesets)
+
+                    # Remove Questions that are no longer on RapidPro.
+                    poll.questions.exclude(ruleset_uuid__in=temba_questions.keys()).delete()
```
dpoirier: Should this be `temba_poll.questions.exclude ...`?

tracpro/polls/views.py
```diff
@@ -191,6 +191,31 @@ def get_form_kwargs(self):
...
+
+                    # Create new or update existing Questions to match RapidPro data.
+                    for order, temba_question in enumerate(temba_questions.values(), 1):
+                        Question.objects.from_temba(poll, temba_question, order)
```
dpoirier: also s/poll/temba_poll/ here?

tracpro/polls/views.py
```diff
@@ -191,6 +191,31 @@ def get_form_kwargs(self):
...
             form.save()
+
+            # Save the associated Questions for this poll here
+            # now that these polls have been activated for the Org
+            selected_polls = []
+            for poll in form.cleaned_data['polls']:
+                selected_polls.append(poll.name)
+
+            org = self.request.org
+            temba_polls = org.get_temba_client().get_flows(archived=False)
```
dpoirier: The rest of this looks very much like PollManager.sync(), except that this only operates on the org's selected polls instead of all of them. Maybe there's some opportunity here to reduce duplication of code?

tracpro/polls/tests/test_models.py
```diff
@@ -172,14 +156,6 @@ def test_sync__update_existing(self):
         self.assertEqual(poll.name, flow.name)
         self.assertTrue(poll.is_active)
 
-        self.assertEqual(models.Question.objects.count(), 1)
-        question = models.Question.objects.get()
-        self.assertEqual(question.poll, poll)
-        self.assertEqual(question.ruleset_uuid, ruleset.uuid)
-        self.assertEqual(question.rapidpro_name, ruleset.label)
-        self.assertEqual(question.name, ruleset.label)
-        self.assertEqual(question.order, 1)
```
dpoirier: Could the test code being deleted instead be rewritten to test the new logic?

tracpro/polls/views.py
```diff
@@ -191,6 +191,31 @@ def get_form_kwargs(self):
 
         def form_valid(self, form):
             form.save()
+
+            # Save the associated Questions for this poll here
+            # now that these polls have been activated for the Org
+            selected_polls = []
+            for poll in form.cleaned_data['polls']:
+                selected_polls.append(poll.name)
```
emullaney: great idea! :unicorn:

tracpro/polls/views.py
emullaney: I guess I see an opportunity to pull out these two lines, I just don't know if I agree it's really worth it...
`temba_polls = org.get_temba_client().get_flows(archived=False)
 temba_polls = {p.uuid: p for p in temba_polls}`

tracpro/polls/views.py
emullaney: weird, I guess it existed in the old code and I didn't realize...

tracpro/polls/views.py
emullaney: yep!

tracpro/polls/views.py
emullaney: argh, good catch.

emullaney:
As @dpoirier and I discovered, the updated task was not updating the questions correctly for the polls that were being synced, which is why the processing was going much faster. I corrected that, moved the sync into a celery task that's called by the form save method so that it can happen in the background. This should prevent the form timeout, on either the form load or the form save.

tracpro/polls/tests/test_forms.py
```diff
@@ -78,3 +79,25 @@ def test_numeric_invalid(self):
             'numeric': ['Select a valid choice. '
                         'invalid is not one of the available choices.'],
         })
+
+
+class TestActivePollsForm(TracProTest):
```
dpoirier: We should also test that the task code is doing what we want.

tracpro/polls/forms.py
```diff
@@ -53,6 +54,21 @@ def save(self):
...
+            selected_poll_names.append(poll.name)
+            selected_polls.append(models.Poll.objects.get(id=poll.id))
+
+        temba_polls = self.org.get_temba_client().get_flows(archived=False)
+        temba_polls = {p.uuid: p for p in temba_polls}
+
+        # Call a celery task to update the questions and categories
+        # This takes a long time, so let's schedule it to run in the background
+        sync_questions_categories.delay(
+            temba_polls, selected_poll_names, selected_polls)
```
dpoirier: I think I'd change the task signature so we just pass it the org and the selected polls:

    sync_questions_categories.delay(self.org, self.cleaned_data['polls'])

then in the task we can call `org.get_temba_client().get_flows(archived=False)` to get the temba_polls (saving another API call on the user thread) and carry on from there.

tracpro/polls/tests/test_forms.py
emullaney: I have added a test for the task specifically. Let me know if you have any more suggestions! Thanks.

tracpro/polls/tests/test_tasks.py
```diff
@@ -0,0 +1,42 @@
...
+        flow_1.uuid = 'abcdefg123'
+        flow_1.name = self.poll_1.name
+        ruleset = mock.Mock()
+        ruleset.uuid = 'goodquestion'
+        flow_1.rulesets = [ruleset]
+        self.mock_temba_client.get_flows.return_value = [flow_1]
+
+        with mock.patch('tracpro.polls.tasks.logger.info') as mock_logger:
+            sync_questions_categories(self.org, self.data)
+            self.assertEqual(models.Question.objects.all().count(), 1)  # only one question should remain
```
dpoirier: No need for `.all()` here.

tracpro/polls/tasks.py
```diff
@@ -130,3 +131,39 @@ def org_task(self, org):
...
+def sync_questions_categories(org, polls):
+    # Create new or update SELECTED Polls to match RapidPro data.
+    from tracpro.polls.models import Poll, Question
+
+    # Save the associated Questions for this poll here
+    # now that these polls have been activated for the Org
+    selected_poll_names, selected_polls = [], []
+    for poll in polls:
+        selected_poll_names.append(poll.name)
+        selected_polls.append(Poll.objects.get(id=poll.id))
```
dpoirier: Since we're passing in the polls that we're interested in, these 4 lines aren't really necessary anymore.

tracpro/polls/tests/test_tasks.py
```diff
@@ -0,0 +1,44 @@
...
+from tracpro.test.cases import TracProTest
+
+from .. import models
+from ..tasks import sync_questions_categories
+
+
+class TestPollTask(TracProTest):
+
+    @mock.patch.object(models.Question.objects, 'from_temba')
+    def test_sync_questions_categories(self, mock_question_from_temba):
```
dpoirier: Can we also test that we create a new local Question if there's a new question in RapidPro?

tracpro/polls/tests/test_tasks.py
```diff
@@ -0,0 +1,44 @@
...
+        flow_1.uuid = 'abcdefg123'
+        flow_1.name = self.poll_1.name
+        ruleset = mock.Mock()
+        ruleset.uuid = 'goodquestion'
+        flow_1.rulesets = [ruleset]
+        self.mock_temba_client.get_flows.return_value = [flow_1]
+
+        # Assert that the 2 questions exist before we sync when one should be deleted
+        self.assertEqual(models.Question.objects.all().count(), 2)
+        with mock.patch('tracpro.polls.tasks.logger.info') as mock_logger:
```
dpoirier: Is there any particular reason to use two different patching techniques (decorator for from_temba, context manager for logger.info)?

tracpro/polls/tests/test_tasks.py
```diff
@@ -0,0 +1,44 @@
...
+            sync_questions_categories(self.org, self.data)
+            self.assertEqual(models.Question.objects.all().count(), 1)  # only one question should remain
```
dpoirier: `.all()` not needed here.

tracpro/polls/tests/test_tasks.py
emullaney: Yes I've added a test line for this, coming soon...

tracpro/polls/tests/test_tasks.py
emullaney: :heavy_check_mark:

tracpro/polls/tests/test_tasks.py
emullaney: no I suppose not, I will simplify...

tracpro/polls/tasks.py
emullaney: Ok, I think I know how to replace this with something a little simpler. But we still need to check that the  poll we get back from temba is in the poll list we want to sync. Let me know if my change makes sense! (coming soon)

tracpro/polls/tasks.py
```diff
@@ -130,3 +131,35 @@ def org_task(self, org):
...
+def sync_questions_categories(org, polls):
+    # Create new or update SELECTED Polls to match RapidPro data.
+    from tracpro.polls.models import Poll, Question
+
+    # Save the associated Questions for this poll here
+    # now that these polls have been activated for the Org
+    selected_poll_names = [poll.name for poll in polls]
+
+    temba_polls = org.get_temba_client().get_flows(archived=False)
+    temba_polls = {p.uuid: p for p in temba_polls}
```
dpoirier: I don't think we need this dictionary any more. (We're just extracting its `.values()`.)

tracpro/polls/tests/test_tasks.py
```diff
@@ -0,0 +1,58 @@
...
+        self.mock_temba_client.get_flows.return_value = [flow_1]
+        # Mock this call to return an empty rule set so that RapidPro API is not called
+        mock_poll_get_flow.return_value.rulesets = []
+
+        # Assert that the 2 questions exist before we sync when one should be deleted
+        self.assertEqual(models.Question.objects.count(), 2)
+
+        # Call the task to sync questions...
+        sync_questions_categories(self.org, self.data)
+        # Two questions exist locallyi, one is new from the RapidPro API mock (flow_1.rulesets)
```
dpoirier: 'locallyi' -> 'locally'?

PR #192, opened 2017-01-18
Change label on login username field to "username" instead of "email"

dpoirier:
since that's what it is expecting


emullaney:
:unicorn: :rainbow:

PR #193, opened 2017-01-25
[TD-2] Add/Edit Superuser via UI

dpoirier:
For superusers, in the "Site Manage -> Users" pages,
allow creating new users as superusers, and toggling
the superuser flag when updating users.


emullaney:
:rabbit: :tophat:

PR #194, opened 2017-02-03
[TD-4] Web interface to fetch past runs

dpoirier:
Added to the org page under Administration - "Fetch runs"
button goes to a form page where user can enter days,
hours, and/or minutes. The fetch is scheduled in the background
and we return to the org page, showing a message that the fetch
has been scheduled.

### DIFF
--- a/tracpro/orgs_ext/tasks.py
+++ b/tracpro/orgs_ext/tasks.py
@@ -6,6 +6,7 @@ import logging
 from celery import signature
 from celery.exceptions import SoftTimeLimitExceeded
 from celery.utils.log import get_task_logger
+from dash.orgs.models import Org
 
 from django.apps import apps
 from django.conf import settings
@@ -13,7 +14,7 @@ from django.core.cache import cache
 from django.core.mail import send_mail
 from django.utils import timezone
 
-from djcelery_transactions import PostTransactionTask
+from djcelery_transactions import PostTransactionTask, task
 
 from temba_client.base import TembaAPIError
 from temba_client.utils import parse_iso8601, format_iso8601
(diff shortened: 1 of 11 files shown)


tracpro/orgs_ext/tasks.py
```diff
@@ -229,3 +230,57 @@ def wrap_logger(self, level, org, msg, *args, **kwargs):
...
+def fetch_runs(org_id, since):
+    """
+    Fetch responses for the org with id=org_id, going back
+    to `since` (datetime).
+
+    Creates or updates Response objects for each run.
+    """
+    from tracpro.polls.models import Poll, Response  # Avoid circular imports
+
+    logger.debug("Running fetch_runs")
```
emullaney: Why log on the debug level?

tracpro/orgs_ext/tasks.py
```diff
@@ -229,3 +230,57 @@ def wrap_logger(self, level, org, msg, *args, **kwargs):
...
+    try:
+        org = Org.objects.get(id=org_id)
+    except Org.DoesNotExist:
+        raise ValueError("No such org with id %d" % org_id)
+    #
+    # if not any([days, hours, minutes]):
+    #     raise ValueError("At least one of days, hours, or minutes must be non-zero")
+    #
+    # since = timezone.now() - relativedelta(minutes=minutes, hours=hours, days=days)
+
```
emullaney: Do we need this commented out code?

dpoirier: Good grief. I review my code so carefully for things like this commented out code before opening a pull request, and still they slip through. Will delete.

dpoirier:
@emullaney one thing that occurs to me is that the user will have no idea if the fetch worked or when it finishes.  If we have email addresses for users, we could email the user who kicked off the fetch?

tracpro/orgs_ext/tasks.py
dpoirier: Well, it's debug rather than, say, info, so it won't clutter up the logs normally. Why log this at all, though... it's not really logging enough to debug anything anyway.  Maybe I'll just delete it.

emullaney:
@dpoirier I think that's a really good idea to email the user. And maybe add a note to the message that says "Scheduled a fetch in the background" like "Scheduled a fetch in the background. You will receive an email when the fetch has completed."

tracpro/orgs_ext/tasks.py
emullaney: Ok

dpoirier:
@emullaney  This has been updated.

tracpro/orgs_ext/tasks.py
```diff
@@ -233,37 +234,42 @@ def wrap_logger(self, level, org, msg, *args, **kwargs):
...
-    #     raise ValueError("At least one of days, hours, or minutes must be non-zero")
-    #
-    # since = timezone.now() - relativedelta(minutes=minutes, hours=hours, days=days)
+
+    # Collect our log messages so we can email them at the end if we want to.
+    messages = []
+
+    def log(s):
+        messages.append(s)
+        logger.info(s)
```
emullaney: clever.

PR #196, opened 2017-02-09
Subclass the TembaClient to add iterability

dpoirier:


dpoirier:
This could use some tests if the approach looks worth pursuing.

PR #197, opened 2017-02-13
API v2 Updates

emullaney:
https://caktus.atlassian.net/browse/TD-3

(See https://github.com/rapidpro/tracpro/pull/195 for more inline comments about these updates)


tracpro/contacts/utils.py
```diff
@@ -0,0 +1,271 @@
...
+        deleted_uuids.append(deleted_incoming.uuid)
+    existing_contacts.filter(uuid__in=deleted_uuids).update(is_active=False)
+
+    return created_uuids, updated_uuids, deleted_uuids, failed_uuids
+
+
+def temba_compare_contacts(first, second, fields=None, groups=None):
+    """
+    Compares two Temba contacts to determine if there are differences. Returns
+    first difference found.
```
dpoirier: It seems like this is intended to compare two versions of the same Temba contact, so maybe the docstring could say something like that.  Also, maybe add a brief description of the `fields` and `groups` arguments' purpose?

tracpro/contacts/utils.py
```diff
@@ -0,0 +1,271 @@
...
+
+def temba_merge_contacts(first, second, mutex_group_sets):
+    """
+    Merges two Temba contacts, with priority given to the first contact
+    """
+    if first.uuid != second.uuid:  # pragma: no cover
+        raise ValueError("Can't merge contacts with different UUIDs")
+
+    # URNs are merged by scheme
+    first_urns_by_scheme = {u[0]: u[1] for u in [urn.split(':', 1) for urn in first.urns]}
```
dpoirier: It looks like this will discard all but the last urn for each scheme.  E.g. if there are two urns starting with 'http:', only the last one will end up in first_urns_by_scheme.  Is that a problem?

Same for the next line.

tracpro/contacts/utils.py
```diff
@@ -0,0 +1,271 @@
...
+    urns_by_scheme = {u[0]: u[1] for u in [urn.split(':', 1) for urn in second.urns]}
+    urns_by_scheme.update(first_urns_by_scheme)
+    merged_urns = ['%s:%s' % (scheme, path) for scheme, path in six.iteritems(urns_by_scheme)]
+
+    # fields are simple key based merge
+    merged_fields = second.fields.copy()
+    merged_fields.update(first.fields)
```
dpoirier: This is absolutely fine.  If you wanted a oneliner, `merged_fields = dict(second.fields, **first.fields)` would also work.

tracpro/contacts/utils.py
```diff
@@ -0,0 +1,271 @@
...
+
+    # first merge mutually exclusive group sets
```
dpoirier: What is this section of code supposed to do, exactly? I'm having a hard time figuring out just from the code whether it's going to do what it's intended to do.

tracpro/msgs/tasks.py
```diff
@@ -55,40 +60,39 @@ def org_task(self, org):
...
 
-        client = org.get_temba_client()
+        def update_or_create_message(message, org):
+            contact = Contact.objects.filter(uuid=message.contact.uuid).first()
+            # If the contact sync task hasn't gotten this contact yet,
+            # don't get the message yet
+            if contact:
+                InboxMessage.objects.update_or_create(
+                    rapidpro_message_id=message.id,
+                    org=org,
```
dpoirier: I believe we want everything after rapidpro_message_id to be passed as the `defaults` dictionary here (https://docs.djangoproject.com/en/1.8/ref/models/querysets/#update-or-create).

tracpro/msgs/views.py
```diff
@@ -170,8 +169,8 @@ def post(self, request, *args, **kwargs):
                 # Wait 1 second to allow message to reach server then run the task
```
dpoirier: This comment about waiting 1 second doesn't appear to apply anymore.  Or maybe the time.sleep(1) needs to be put back in?

tracpro/polls/migrations/0030_reset_question_types.py
```diff
@@ -21,7 +20,8 @@ def guess_question_type_from_rules(apps, schema_editor):
     for org in apps.get_model('orgs', 'Org').objects.all():
         client = get_temba_client(org)
         for poll in org.polls.all():
-            rulesets = client.get_flow_definition(poll.flow_uuid).rule_sets
+            exports = client.get_definitions(flows=[poll.flow_uuid])
+            rulesets = exports.flows[0].rule_sets
```
dpoirier: This would be a later optimization, but it looks like the v2 API would let us fetch the rulesets for all the polls at once (by passing all the flow uuid's in `flows`). Should we add a FIXME comment or an issue or task or something to look into that later?

tracpro/polls/tasks.py
```diff
@@ -48,8 +47,8 @@ def org_task(self, org):
 
         total_runs = 0
         for poll in Poll.objects.active().by_org(org):
-            poll_runs = client.get_runs(flows=[poll.flow_uuid], after=last_time, before=until)
-            total_runs += len(poll_runs)
+            poll_runs = client.get_runs(flow=poll.flow_uuid, after=last_time, before=until)
+            total_runs += len(poll_runs.all())
```
dpoirier: I'm not sure about this `.all()` - do we need it here?  Our client makes the return value from client calls iterable without calling `.all()`.

tracpro/contacts/models.py
```diff
@@ -178,23 +190,20 @@ def get_urn(self):
     @classmethod
     def kwargs_from_temba(cls, org, temba_contact):
         """Get data to create a Contact instance from a Temba object."""
-
         def _get_first(model_class, temba_uuids):
             """Return first obj from this org that matches one of the given uuids."""
             queryset = model_class.get_all(org)
             tracpro_uuids = queryset.values_list('uuid', flat=True)
+            temba_uuids = [temba_uuid.uuid for temba_uuid in temba_uuids]
```
emullaney: Yes that makes sense, will do

tracpro/contacts/signals.py
```diff
@@ -41,9 +41,9 @@ def set_groups_to_new_contact(sender, instance, created, **kwargs):
     """Hook to set the groups of a temba contact when a Contact is created after sync."""
     if created:
         try:
-            temba_contact = instance.org.get_temba_client().get_contact(uuid=instance.uuid)
-            for group_uuid in temba_contact.groups:
-                instance.groups.add(Group.objects.get(uuid=group_uuid))
-        except TembaNoSuchObjectError:
+            temba_contact = get_client(instance.org).get_contacts(uuid=instance.uuid)[0]
+            groups = Group.objects.filter(uuid__in=temba_contact.groups)
```
emullaney: Yes that is a good call.

tracpro/polls/tasks.py
emullaney: This fails without the .all(). I was getting a failure from the celery task yesterday when I pushed this out to staging. The .all() isn't needed elsewhere but it is needed to calculate the len().

tracpro/contacts/utils.py
emullaney: Agreed. I've updated this doc string with all of this information.

tracpro/contacts/utils.py
emullaney: This is okay/intended. This has not been modified from the original code we were using (in dash). I just pulled it down locally.

emullaney: I just stepped through the code and this works ok. We only save one urn per contact on TracPro historically. Temba's (RapidPro's) urns are left unchanged.

tracpro/contacts/utils.py
emullaney: This section of code was copied in from dash (it was removed from dash). 
I believe it is merging the groups that are in each of the two contacts it's creating and creating a list of mutually exclusive groups.

tracpro/msgs/tasks.py
emullaney: Not sure why to use the defaults dictionary on all of those, if some of those fields are required (ie contact). Can you explain?

tracpro/msgs/views.py
emullaney: No I left the comment by mistake. We no longer need the sleep at least from my functional testing. I will remove it.

tracpro/polls/migrations/0030_reset_question_types.py
emullaney: Good call. Adding FIXME/TODO.

tracpro/msgs/tasks.py
emullaney: I've updated this, using the `defaults` dictionary for the non-required fields.

dpoirier:
:railway_track:

PR #201, opened 2017-02-28
Updates for deploy for 1.5.1

emullaney:
* API v2 Updates: Code now points to version 2 of the RapidPro API.


tracpro/client.py
```diff
@@ -0,0 +1,53 @@
...
+        return TracProCursorQuery(self, '%s/%s.json' % (self.root_url, endpoint), params, clazz)
+
+    def get_boundaries(self):
+        return self._get_query('boundaries', {'geometry': 'true'}, Boundary)
+
+
+class TracProCursorQuery(CursorQuery):
+    """
+    Customized CursorQuery that allows iterating over it without having
+    to call `.all()` on it. More Pythonic.
```
rowanseymour: Each to his own :) The `all()`, `first()` methods are obviously mimicing Django querysets. It's also important to note that `all()` fetches all pages behind the scenes, but there are often cases when you want to explicitly handle each request/page at a time, e.g. updating local state as you go.

dpoirier: Right, this is definitely customized for our particular use case and not intended to be a fully general solution.

Thought I might note that you can also iterate over a Django queryset without having to call `.all()`.

PR #202, opened 2017-03-01
[TD-18] Fixes for error emails from tasks

dpoirier:
* Turn off 'fail_silently' so we might see problems with sending
  emails in the logs, at least
* Change the name of our send_error_email method. It was overriding
  another method of the same name but different arguments from the
  parent class, so when the parent class tried to send an error email,
  it ended up in our method and didn't work.


emullaney:
:rainbow:

PR #203, opened 2017-03-03
TD-20 staff permissions on fetch runs front end

emullaney:
I am hoping that either @rowanseymour OR @norkans7 can help me figure out why this permission in smartmin is not allowing "administrators" access to the form in question. It's a small pull request. I'm hoping one of you might be able to guide me in the right direction?

### DIFF
--- a/tracpro/settings/base.py
+++ b/tracpro/settings/base.py
@@ -271,6 +271,7 @@ GROUP_PERMISSIONS = {
     "Administrators": (
         'orgs.org_home',
         'orgs.org_edit',
+        'orgs.fetch_runs',
         'baseline.baselineterm.*',
         'contacts.contact.*',
         'groups.boundary.*',
--- a/tracpro/settings/base.py
+++ b/tracpro/settings/base.py
@@ -271,7 +271,7 @@ GROUP_PERMISSIONS = {
     "Administrators": (
         'orgs.org_home',
         'orgs.org_edit',
-        'orgs.fetch_runs',
+        'orgs.org_fetch_runs',
         'baseline.baselineterm.*',
         'contacts.contact.*',
         'groups.boundary.*',
--- a/tracpro/settings/base.py
+++ b/tracpro/settings/base.py
@@ -333,7 +333,7 @@ PERMISSIONS = {
         'delete',  # can delete an object
         'list',  # can view a list of the objects
     ),
-    'orgs.org': ('create', 'update', 'list', 'edit', 'home'),
+    'orgs.org': ('create', 'update', 'list', 'edit', 'home', 'fetch_runs'),
     'baseline.baselineterm': ('create', 'read', 'update', 'delete', 'list', 'data_spoof', 'clear_spoof'),
     'contacts.contact': ('create', 'read', 'update', 'delete', 'list'),
     'groups.group': ('list', 'most_active', 'select'),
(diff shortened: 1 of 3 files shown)


rowanseymour:
I think you need to add it as a new permission in `settings.PERMISSIONS`

rowanseymour:
Oh and then also run `manage.py migrate` to add it as a permission in the db

tracpro/settings/base.py
```diff
@@ -271,6 +271,7 @@ def _org_scheduler_task(task_name):
     "Administrators": (
         'orgs.org_home',
         'orgs.org_edit',
+        'orgs.fetch_runs',
```
emullaney: @rowanseymour is this not the way to add it to the permissions? I have been running `migrate` but am not getting any feedback that it's doing anything and it doesn't seem to work. Any other input is welcome!

rowanseymour: That's the way of assigning permissions to groups - adding them to PERMISSIONS is the way of adding them to the database

emullaney: Oh I did miss that step. Thanks! I'll try that.

tracpro/settings/base.py
```diff
@@ -333,7 +333,7 @@ def _org_scheduler_task(task_name):
         'delete',  # can delete an object
         'list',  # can view a list of the objects
     ),
-    'orgs.org': ('create', 'update', 'list', 'edit', 'home'),
+    'orgs.org': ('create', 'update', 'list', 'edit', 'home', 'fetch_runs'),
```
emullaney: @rowanseymour `python manage.py migrate` still has no feedback after adding this to the PERMISSIONS setting. And going to this page as an administrator still brings me to the login page. 

Am I missing something else? Does this permission work on SmartFormView?

rowanseymour: Ahh I remember running into this previously - what you've got here creates a permission called `orgs.org_fetch_runs`. Smartmin insists on checking that there's an app called `orgs` and a model called `org`. 

So I think you have to use `orgs.org_fetch_runs` as your permission on the view

emullaney: Tried that. Still no dice. Pushing that up now... :/

rowanseymour: Can you look in the database after doing the `migrate` to see if the permission is created and added to the Administrators group?

emullaney: @rowanseymour yes, it is there:
`Group.objects.get(name='Administrators').permissions.filter(content_type=content_type)`
`[<Permission: orgs | org | Can edit org>, <Permission: orgs | org | Can fetch_runs org>, <Permission: orgs | org | Can home org>]`

emullaney: So I'm thinking that because the permission is there and it was created during one of my migrations, there might be some issue with smartmin recognizing it when I try to go to that form page as an administrator?

rowanseymour: Try making the view extend `OrgPermsMixin`

emullaney: Oh my gosh that was it, Rowan.

emullaney: :medal_sports:

emullaney:
@dpoirier this is ready for review

dpoirier:
I checked out the branch and tried it, but the permission (org.org_fetch_runs) does not exist. Do we need to add a migration or something, so this'll get created when we deploy it?

emullaney:
@dpoirier you have to run `python manage.py migrate` so that smartmin creates the new permission... do you think that will be a problem with our deploy process? Can you try that locally?

dpoirier:
Okay, it looks like just running a migrate, even without adding any new migrations, gets the permission created.  Our deploy should do that every time, so we should be good.

🎩

Issue #3, opened 2015-03-25
configurable label for geographic areas


ewheeler:
Currently, any geographic organizations of tracpro users are labelled as `regions`

This is confusing in scenarios where a country has an official administrative level called 'region' but tracpro is configured using a different official administrative level, such as 'districts'

nicpottier:
I think we use ```state``` and ```district``` are our level 1 and level 2 names.

I feel this one is slightly tricky, as yes, obviously every country calls these different things, but then how do we refer to them in documentation, in dropdowns etc?

ewheeler:
Very tricky. Using `state`, `district`, and `region` (which often have very precise meanings) as generic terms is very confusing without additional explanation for users. I think using deliberately generic terms like `admin1` and `admin2` would lessen confusion overall. 

The case for tracpro is a bit different, because geographic areas can really be arbitrary as it just represents a group of users (that may or may not have some geographic affinity -- could be organizational divisions or ministry departments etc). So for tracpro it could even be free text entered by the admin.

ewheeler:
Another thought: for tracpro it might make sense to generalize even further -- and have them be `cohorts` instead of having any explicit geographic anchor. Then maybe have a cohort type that is geographic (so a 'district' cohort can be associated with country x's admin2 level) that could then be displayed on the currently non-existent map..

ewheeler:
`panel` is the term used in the market research industry to refer, more or less, to "a group of people we contact regularly (and know their demographics) that we can send surveys to"

kindof obscure but perhaps more precise. 

`cohort` still might make more sense for our users, but wanted to throw this in the ring too

daaray:
:+1: for `cohort` in this context.

rowanseymour:
TracPro is currently modelling two types of cohort and I wonder if naming shouldn't distinguish them. There is:
  * __region__ - a cohort of contacts (typically) associated by geography. Supervisor users are assigned permissions to particular cohorts.
  * __reporting group__ - a cohort of contacts associated by role (e.g. Males, Teachers, Farmers)

`cohort` is certainly an apt word for the former... but that would seem to be make the latter less clear

ewheeler:
would it make sense for them to be 
`cohort` (region)
`role` (reporting group)
?

rowanseymour:
[...]

PR #204, opened 2017-03-07
Rename regions to panels and reporter groups to cohorts

dpoirier:
See https://github.com/rapidpro/tracpro/issues/3
Name changes:
Regions -> Panels
Reporter Groups -> Cohorts

### DIFF
--- a/tracpro/polls/views.py
+++ b/tracpro/polls/views.py
@@ -484,7 +484,10 @@ class ResponseCRUDL(smartmin.SmartCRUDL):
 
     class ByPollrun(OrgPermsMixin, smartmin.SmartListView):
         default_order = ('-updated_on',)
-        field_config = {'updated_on': {'label': _("Date")}}
+        field_config = {
+            'updated_on': {'label': _("Date")},
+            'region': {'label': _('Panel')},
+        }
         link_fields = ('contact',)
 
         def dispatch(self, request, *args, **kwargs):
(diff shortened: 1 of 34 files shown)


tracpro/polls/views.py
```diff
@@ -600,7 +600,10 @@ def render_to_response(self, context, **response_kwargs):
                 questions = self.derive_questions().values()
 
                 resp_headers = ['Date']
-                contact_headers = ['Name', 'URN', 'Region', 'Group']
+                # FIXME: Should we change the column name "Region" to "Panel" and possibly break
+                # existing integrations that work by exporting .CSVs?
+                # Or leave it as "Region", and confuse people trying to create new integrations?
```
emullaney: Dan, can we hipchat about this one whenever you have time? No rush.

emullaney: The short answer is that I'm not concerned about existing integrations (I doubt that there are any to worry about). I would rather not confuse people, I would suggest we use the new name here.

emullaney:
Caught one missing change on this page: http://caktus.localhost:8000/pollrun/participation/10/?group-by=region
image

Wording on the page says "Region" on the column header. Not concerned about the url.

emullaney:
"Region" should be "Panel" here on this column header as well, which I think I mentioned already in regards to a FIXME comment: http://caktus.localhost:8000/response/by_pollrun/13/

emullaney:
@dpoirier this is ready for some suggested fixes. I've reviewed it in it's entirety. Looks pretty great so far.

dpoirier:
@emullaney This has been updated and is ready for re-review.

emullaney:
:sunflower:

PR #205, opened 2017-03-08
[TD-13] Fix redirect loop on change password

dpoirier:
If a user has the change_password flag set, and tries to visit a
non-org URL, they should not get stuck in a redirect loop. Instead,
make them land on the "you need to pick an org" page.


emullaney:
:cactus:

PR #206, opened 2017-03-08
[TD-16] Fixes for admins

dpoirier:
* When someone is added to an org as an admin, make sure they're
  flagged as staff.
* In the users list, add a column showing whether a user is staff or
  not.


emullaney:
:1st_place_medal:

PR #207, opened 2017-03-15
[TD-12] Create smart categories on single date numeric charts

dpoirier:


dpoirier:
@emullaney Please take a look. I think this basically does what we wanted, but would be happy to tweak the details. Also, I'd like to be sure this isn't breaking other charts than single pollrun page I looked at it on.

tracpro/polls/models.py
```diff
@@ -666,7 +671,46 @@ def category_counts(self):
...
+        can't be successfully converted to a float.
+
+        Returns dictionary {
+          'categories': list of category names in order,
+          'data': list of counts in order
+        }
+
+        Category names are of the form "N.N-N.N".
+        """
+        answers = just_floats(self.filter(category='numeric').values_list('value', flat=True))
```
emullaney: I think what we actually want to filter on here is `self.filter(question__question_type=Question.TYPE_NUMERIC)`:

`answers = just_floats(self.filter(question__question_type=Question.TYPE_NUMERIC).values_list('value', flat=True))`

tracpro/polls/models.py
```diff
@@ -666,7 +671,46 @@ def category_counts(self):
...
+        if not answers:
+            return {
+                'categories': [],
+                'data': [],
+            }
+
+        hist, bin_edges = np.histogram(answers, bins='sqrt')
+
+        category_names = ["%r-%r" % (bin_edges[i], bin_edges[i+1]) for i in range(len(hist))]
```
emullaney: Perhaps this change will make it more readable on the chart:
`category_names = ["%r-%r" % (round(bin_edges[i], 2), round(bin_edges[i+1], 2)) for i in range(len(hist))]`

tracpro/polls/tests/test_charts.py
```diff
@@ -247,6 +247,8 @@ def test_single_pollrun_open(self):
         self.assertEqual(summary_table, None)
 
     def test_single_pollrun_numeric(self):
+        # Make answers numeric
+        models.Answer.objects.update(category='numeric')
```
emullaney: We'll want to update this test per an earlier comment to update the Questions to numeric.

dpoirier:
@emullaney  Updated and ready for re-review.

emullaney:
:cat2: :cat2:

PR #208, opened 2017-03-17
Updates for deploy for 1.5.2

emullaney:
* Larger, sorted multi-select drop-downs
* Fixes for error emails from tasks
* Permissions fix for fetch runs front-end: Admins can now use this feature, in addition to super users.
* Fix redirect loop on change password
* Fixes for admins:
    -  When someone is added to an org as an admin, make sure they're flagged as staff.
    -  In the users list, add a column showing whether a user is staff or not.


dpoirier:
+1

PR #210, opened 2017-04-06
Fixes to contact syncing

dpoirier:
* Get all rapidpro contacts for the selected groups,
  rather than just ones modified more recently than
  some time. This should make sure we get all the
  relevant contacts.

* Add new SyncPullTest testcase, which is intended
  to check out all the different cases in which
  something has changed on RapidPro and we then
  sync contacts.

Also, simplify the code in some places, e.g.
remove a bunch of parameters from sync_pull_contacts
that we were always passing in the same value for
anyway.

### DIFF
--- a/tracpro/contacts/models.py
+++ b/tracpro/contacts/models.py
@@ -215,13 +215,11 @@ class Contact(models.Model):
         # Use the first Temba group that matches one of the org's Regions.
         region = _get_first(Region, temba_contact.groups)
         if not region:
-            raise NoMatchingCohortsWarning(("Unable to save contact {c.uuid} ({c}) because none of "
-                                            "their cohorts ({cohorts}) match an active Panel "
-                                            "({panels}) for this org.").format(
-                c=temba_contact,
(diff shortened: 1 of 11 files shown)


tracpro/contacts/models.py
```diff
@@ -191,19 +208,20 @@ def get_urn(self):
...
         # Use the first Temba group that matches one of the org's Regions.
         region = _get_first(Region, temba_contact.groups)
         if not region:
-            raise ValueError(
-                "Unable to save contact {c.uuid} ({c.name}) because none of "
-                "their cohorts match an active Panel for this org.".format(
-                    c=temba_contact))
+            raise NoMatchingCohortsWarning(("Unable to save contact {c.uuid} ({c}) because none of "
+                                            "their cohorts ({cohorts}) match an active Panel "
+                                            "({panels}) for this org.").format(
```
emullaney: We talked about this in chat already, but let's make this more readable by using including the contact's name and excluding `{cohorts}` and `{panels}`

PR #211, opened 2017-04-06
Add Google Analytics support

dpoirier:
Org admins can add a google analytics tracking
code to the org settings, and the org's pages
will be tracked in GA.


dpoirier:
@emullaney Sorry, I missed seeing that you had asked for a change on this. I've now updated it and it's ready for another look.

emullaney:
:rainbow: :sunflower:

PR #214, opened 2017-04-12
Skip extra groups query for new contacts

emullaney:
Signal to make fewer calls to RapidPro API and improve performance of the contact sync


emullaney:
Opened by @dpoirier and approved once travis completes successfully :kiwi_fruit:

PR #216, opened 2017-04-24
[TD-23] Adding logic to filter out polls that begin with the words 'Single message' regardless of case.

emullaney:


tracpro/polls/models.py
```diff
@@ -71,8 +71,13 @@ def set_active_for_org(self, org, uuids):
...
-        temba_polls = {p.uuid: p for p in temba_polls}
+        temba_polls_result = get_client(org).get_flows()
+
+        # Filter out polls with names starting with 'Single Message'
+        temba_polls = {}
+        for poll in temba_polls_result:
+            if poll.name.lower()[:14] != 'single message':
+                temba_polls[poll.uuid] = poll
 
         # Remove Polls that are no longer on RapidPro.
```
dpoirier: Maybe add to this comment, ", or that we're filtering out".?

tracpro/polls/models.py
```diff
@@ -71,8 +71,13 @@ def set_active_for_org(self, org, uuids):
...
         """Update the org's Polls from RapidPro."""
         # Retrieve current Polls known to RapidPro.
-        temba_polls = get_client(org).get_flows()
-        temba_polls = {p.uuid: p for p in temba_polls}
+        temba_polls_result = get_client(org).get_flows()
+
+        # Filter out polls with names starting with 'Single Message'
+        temba_polls = {}
+        for poll in temba_polls_result:
+            if poll.name.lower()[:14] != 'single message':
```
dpoirier: I only see "Single Message" in the examples, is there any need to make this case-insensitive?  Maybe we could just say ```if not poll.name.startswith("Single Message")```?

tracpro/polls/models.py
emullaney: Sure.

tracpro/polls/models.py
emullaney: poll isn't a Django queryset object but rather a tembaclient Flow object, so startswith isn't available. 

I could remove the lower() if we don't think it's necessary, that's one less string operation...

dpoirier: startswith is a Python string operation, so should be available on any string.

emullaney: Oh cool! I didn't know that one. Just added it.

dpoirier:
+1

PR #217, opened 2017-04-24
Skip some unneeded RapidPro queries during sync

dpoirier:
When creating a new contact during a sync, we already know the new
contact's groups and don't need to query RapidPro for them.


emullaney:
:rocket:

PR #220, opened 2017-04-26
Use different Redis databases for celery, cache

dpoirier:
On deployed servers, we were putting the celery
broker, the results, and the cache all in the
same Redis database. When we cleared the cache,
we clobbered our Celery queue.

Put each one in its own database so this shouldn't
happen again.


emullaney:
:champagne:

PR #221, opened 2017-05-04
Rearrange top menu bar

dpoirier:


emullaney:
I've updated the AC in the original ticket and that will require some changes:

- Super users have access to all the items in the list, except "My Organization" which does not display.
- Admin users (users who were created by selecting them from the multi-select box on the Organization update page) have access to everything *except* "Site Mgmt" items. They also see "My Organization" which links them to organization detail for their organization (ie. Caktus or Sierra Leone). They are the only users that see that.
- Supervisors have access to Contacts, Inbox Messages and Recent Indicators only

dpoirier:
@emullaney  Updated again.

PR #226, opened 2017-05-05
[DIS-28] Contact sync fixes with more logging

emullaney:


dpoirier:
I was doing something similar on a local branch in even more detail, if you want to see I've now pushed it: https://github.com/rapidpro/tracpro/compare/debugging_contact_syncing?expand=1

It might not be useable as-is, I was only intending it as a temporary local change to see how things worked.

dpoirier:
+1

PR #236, opened 2017-05-10
DIS-28: Always update contacts

emullaney:
Update contacts regardless of whether there is a diff. Because we want to ensure we capture their most current group relationships and that will not always be captured by the diff. 
Update tests accordingly.


dpoirier:
:shipit:

PR #238, opened 2017-05-15
[TD-35] [TD-37] [TD-38] Updates to text for panels and cohorts pages.

emullaney:
Clarifying the text on the panel and cohorts forms and list pages for improved UX.


dpoirier:
+1

PR #240, opened 2017-05-23
Two sync problems

dpoirier:
1) When processing a synced response, any contact
showing up that was not in a selected cohort would
abort that response's sync (maybe more).

2) It appears from some logged exceptions that we
sometimes get back from RapidPro flow definitions
that we didn't ask for. Handle that case graceully
by skipping the unwanted definition.

### DIFF
--- a/tracpro/polls/tasks.py
+++ b/tracpro/polls/tasks.py
@@ -144,6 +144,7 @@ def sync_questions_categories(org, polls):
     # now that these polls have been activated for the Org
     flow_uuids = [poll.flow_uuid for poll in polls]
     total_polls = len(flow_uuids)
+    polls_by_uuid = {poll.flow_uuid: poll for poll in polls}
 
     logger.info(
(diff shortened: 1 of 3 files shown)


tracpro/polls/tasks.py
```diff
@@ -144,16 +144,26 @@ def sync_questions_categories(org, polls):
...
-        poll = polls[flow_uuids.index(flow['metadata']['uuid'])]
+        uuid = flow['metadata']['uuid']
+        if uuid not in flow_uuids:
+            logger.error("Ignoring flow definition with uuid %r that we didn't ask for. "
+                         "We asked for %r." % (uuid, flow_uuids))
+            continue
+
+        poll = polls_by_uuid[uuid]
+
+        logger.info("Retrieving questions and categories for flow %s (%s)" % (poll.flow_uuid, poll.name))
```
emullaney: Nice work adding the logging!

PR #243, opened 2017-05-26
[TD-29] infographic wizard

emullaney:
New infographic wizard on the Panel, Cohort, Poll and Fetch Run Pages

Follows UX mockup from https://caktus.atlassian.net/browse/TD-29 but the order has been changed to the above order (because you need contacts before you can track polls). The Jira ticket has been updated as well.

### DIFF
--- a/tracpro/static/less/trac.less
+++ b/tracpro/static/less/trac.less
@@ -231,7 +231,6 @@ footer {
 }
 
 .circle-label {
-
     font-size: .5em;
     color: #000000;
     line-height: 1.5em;
--- a/tracpro/static/less/trac.less
+++ b/tracpro/static/less/trac.less
@@ -204,3 +204,14 @@ footer {
 .ui-datepicker select.ui-datepicker-month, .ui-datepicker select.ui-datepicker-year {
   color: #000000;
 }
+
+/* Workflow Tracker Styles */
+.circle {
+  border-radius: 50%;
+  width: 75px;
+  height: 75px;
+  background-color: #2A5032;
+  text-align: center;
+  color: #FFFFFF;
+  font-size: 2em;
+}
--- a/tracpro/static/less/trac.less
+++ b/tracpro/static/less/trac.less
@@ -206,12 +206,42 @@ footer {
 }
 
 /* Workflow Tracker Styles */
-.circle {
+.workflow-tracker {
+  margin-left: 150px;
+  margin-bottom: 55px;
+}
+
+.circle, .circle-selected {
   border-radius: 50%;
   width: 75px;
   height: 75px;
-  background-color: #2A5032;
+  background-color: #CCC;
   text-align: center;
   color: #FFFFFF;
   font-size: 2em;
+  line-height: 2em;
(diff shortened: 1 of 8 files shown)


tracpro/templates/workflow_tracker_steps.html
```diff
@@ -0,0 +1,40 @@
...
+        <div class="col-md-2 col-sm-1 hidden-xs {% if step == 3 %}circle-selected{% else %}circle{% endif %}">
+            3
+            <div class="circle-label">
+                Select Flows
+            </div>
+        </div>
+    </a>
+    <div class="col-md-2 hidden-sm hidden-xs workflow-spacer"></div>
+
+    <a href="{% url 'orgs_ext.org_fetchruns' %}">
```
dpoirier: We should only display the 4th circle if the user is authorized to do fetchruns.

(```  {% if request.user.is_superuser or request.user.is_staff %}``` is what we've used elsewhere.)

tracpro/static/less/trac.less
```diff
@@ -203,4 +203,44 @@ footer {
 /* Override datepicker colors */
 .ui-datepicker select.ui-datepicker-month, .ui-datepicker select.ui-datepicker-year {
   color: #000000;
```
dpoirier: When I narrow my window, the circles and text end up scrunched on top of each other while there's still horizontal space that they could spread out in.

image

tracpro/static/less/trac.less
```diff
@@ -203,4 +203,44 @@ footer {
...
-}
\ No newline at end of file
+}
```
dpoirier: I noticed some differences from the mockup:

* On the mockup, the circles had a very light circle drawn around them just a few pixels outside the colored circle.
* The text in the unselected circles was dark rather than light.
* The labels were bold.
* The circles appear smaller in the mockup, though I haven't measured them.

tracpro/templates/workflow_tracker_steps.html
emullaney: Actually that should be true of the entire wizard! I will wrap it.

emullaney: I believe that staff and super users should be able to get to any of these pages so that's why I didn't include it to begin with.

tracpro/static/less/trac.less
emullaney: I'm fixing the scrunching by hiding the words on small screen sizes. Fix coming.

tracpro/static/less/trac.less
emullaney: I've made a few improvements here. I think it's close enough to the mockup now. Let me know!

emullaney:
I have
* Updated the styles for the workflow tracker
* Added to the workflow tracker to My Organization
* Fixed the drop down order in the navigation to match the new tracker

Ready for review again @dpoirier

dpoirier:
Things look okay to me, just a little different stylistically from the mockup. Your call if you want to :shipit:.

PR #246, opened 2017-07-05
update message files for translation

ewheeler:
looks like the [po files](https://github.com/rapidpro/tracpro/tree/develop/locale) haven't been updated since Nov 2015, so here are updates per the [docs](https://tracpro.readthedocs.io/en/latest/developers/translation.html#update-translations)


dpoirier:
Looks reasonable.

PR #248, opened 2017-08-29
Minor copy-editing of docs

dpoirier:


docs/users/technical-overview.rst
```diff
@@ -5,10 +5,10 @@ Concepts
 --------------------
 
 * `Polls` or `Flows` are RapidPro flows. Click on the name of a poll to see a summary of the responses over time and dates it was conducted.
-* `PollRuns` are the dates a poll was conducted. Click on a poll date to see a summary of just the responses for that date, participation among cohorts, send a message to contacts, individual responses, and export responses.
-* `Contacts` are the same as in RapidPro, they are the respondents of polls.
-* `Supervisors` could be, for example, a provincial monitoring officer in a ministry. They are meant to see a particular panel or multiple panels, and encourage participation by seeing who is responding, and restart polls. Supervisors also manage contacts.
-* `Administrators` setup TracPro for an org and can manage everything supervisors can, in addition to managing supervisors.
+* `PollRuns` are the dates a poll was conducted. Click on a poll date to see a summary of just the responses for that date, see participation among cohorts, send a message to contacts, see individual responses, and export responses.
```
philtuskha: The docs say `PollRuns` here, but the headline of the main list pages reads `All Flow Runs`.  The Main nav also says 'Flow Management' and in the dropdown, there are items, `Flows` and `FlowRuns` .   Depending on what we want here, We should either...
1) change this "PollRuns" to "FlowRuns" and call it a day.
2) add something to the effect of "PollRuns (or FlowRuns in RapidPro)...." here, and then change all mentions of FlowRuns in TracPro to PollRuns.  Also, have main navigation category read, "Poll Management."   

option 2 would possibly be a seperate pr

philtuskha: `see participation among cohorts` could change to `see participation among cohorts and panels`

ewheeler: i'd vote for option 1. don't see much use for using a synonym rather than being consistent with rapidpro's names

dpoirier:
Updated.

PR #249, opened 2017-09-07
[TP-54] Only allowing one response per contact per day

philtuskha:


tracpro/polls/tests/test_views.py
```diff
@@ -113,3 +113,55 @@ def test_by_contact(self):
...
+            status=Response.STATUS_COMPLETE)
+
+        self.pollrun1_r2 = factories.Response(
+            pollrun=self.pollrun, contact=self.contact1,
+            created_on=dates[1], updated_on=dates[1],
+            status=Response.STATUS_COMPLETE)
+
+        self.pollrun1_r3 = factories.Response(
+            pollrun=self.pollrun, contact=self.contact1,
+            created_on=dates[2], updated_on=dates[0],
```
dpoirier: looks like a typo here, it's setting updated_on to the first date instead of the third.

tracpro/polls/tests/test_views.py
```diff
@@ -113,3 +113,55 @@ def test_by_contact(self):
...
+            status=Response.STATUS_COMPLETE)
+
+        # Adding Bob response on the same exact date as an Ann response
+        self.pollrun1_r4 = factories.Response(
+            pollrun=self.pollrun, contact=self.contact2,
+            created_on=dates[2], updated_on=dates[0],
```
dpoirier: here too?

tracpro/polls/models.py
```diff
@@ -472,7 +472,21 @@ def get_responses(self, region=None, include_subregions=True,
                 responses = responses.filter(contact__region=region)
         if not include_empty:
             responses = responses.exclude(status=Response.STATUS_EMPTY)
-        return responses.select_related('contact')
+
+        responses = responses.select_related('contact')
+
+        # Remove duplicate resopnse from the same user on the same day
+        response_tuple_list = []
```
dpoirier: ``response_tuple_list`` would be a tiny bit more efficient as a set, since we don't care about order or duplicates, and we're checking for membership in it a lot.

tracpro/polls/models.py
```diff
@@ -472,7 +472,21 @@ def get_responses(self, region=None, include_subregions=True,
                 responses = responses.filter(contact__region=region)
         if not include_empty:
             responses = responses.exclude(status=Response.STATUS_EMPTY)
-        return responses.select_related('contact')
+
+        responses = responses.select_related('contact')
+
+        # Remove duplicate resopnse from the same user on the same day
+        response_pk_list, response_tuple_set = [], set()
```
dpoirier: Initializing several variables in one line this way is valid Python, but my feeling is that readability suffers, and it's generally better to put each one on its own line.

Which is not to say this kind of thing is never appropriate. It's great for breaking out values from a tuple or list into separate variables, or atomically swapping values.

dpoirier: My tendency would be to use variable names like "response_pks" and "response_tuples" rather than building the Python type into the variable name. Again, for readability.

This one is fuzzier, though. There are definitely times when I'd want to be explicit in the variable names about what types things are, to avoid confusion or even possible errors when using them. And this case, where we have two collections being used in the same code that are of different types, might be such a time.  Though maybe we could have both readability and less chance of forgetting what type things are if we just made both collections sets here... the collection of PKs doesn't need to be a set, but it wouldn't really hurt.  Sometimes it's worth sacrificing a micro-optimization for readability and reducing the chance of someone accidentally breaking the code when making later changes.

To summarize ... I'd like something done here so we end up with better variable names, but there are multiple options I'd be okay with to address it.

PR #250, opened 2017-09-15
[TD-58] Let orgs optionally sum multiple responses

dpoirier:
More specifically, let orgs configure an optional
alternate behavior when the same contact answers the
same numeric question more than once on the same day.
Ordinarily we use the last answer submitted, but they
can instead choose to sum the numeric responses.

This is an org-wide option, set in the Org's configuration
page that is only accessible to org administrators
(or maybe to superusers? something like that).

### DIFF
--- a/tracpro/polls/models.py
+++ b/tracpro/polls/models.py
@@ -783,12 +783,12 @@ class Answer(models.Model):
 
     def save(self, *args, **kwargs):
         is_new = self.pk is None
-        if is_new:
-            self.value_to_use = self.compute_value_to_use()
+        if is_new and not self.should_use_sum():
+            self.value_to_use = self.value
         super(Answer, self).save(*args, **kwargs)
         if is_new and self.should_use_sum():
-            # We created a new answer - we might need to update the summed
-            # values in other answers
+            # We created a new answer - we need to update the summed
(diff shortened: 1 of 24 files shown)


dpoirier:
FYI: I noticed while working on this that Poll.get_flow_definition() was unused (apart from tests), and so removed it.

tracpro/polls/models.py
```diff
@@ -624,12 +621,22 @@ def from_run(cls, org, run, poll=None):
...
+            # If more than one for this contact and pollrun,
+            # set the last one created as the active one.
+            if Response.objects.filter(pollrun=pollrun, contact=contact).count() > 1:
+                # Set them all False, then the one we want true
+                with transaction.atomic():
+                    Response.objects.filter(pollrun=pollrun, contact=contact).update(is_active=False)
+                    last = Response.objects.filter(pollrun=pollrun, contact=contact).order_by('-created_on').first()
+                    last.is_active = True
+                    last.save()
+
```
philtuskha: 👍 this feels better!

dpoirier:
RapidPro was having some uptime issues yesterday so that might be the cause of your problems.

I've updated this pull request.

PR #251, opened 2017-09-15
 [TD 42-45] sync multiple cohorts

philtuskha:
Removed `group` from Contact model.  Using `groups` where we used to use just one. All tests related to Models and Utils (syncing) pass.  

TO DO:
Work on Views and Forms.


philtuskha:
Things to note:
- I was getting an error with `contacts.tests.ContactCRUDLTest.test_create_with_access` but seemingly no problem with actually creating a new contact on localhost.  I decided to skip this test to move these tickets along.    

- in `groups.models.AbstractGroup`, when reviewing, one should probably look at `get_response_counts()`.  For Groups, we are not actually getting response counts here since there is no longer one group per contact and therefore per response.  However, it appears to me that in the only place this function is used for Groups, we are trying to find the most responses per group.  Perhaps refactoring here is in order.  Hopefully, overall, my thinking here is sound, but a second pair of eyes on this would be helpful.

tracpro/contacts/tests/test_models.py
```diff
@@ -113,29 +111,31 @@ def test_kwargs_from_temba(self):
         )
 
         kwargs = models.Contact.kwargs_from_temba(self.unicef, temba_contact)
+
+        # try creating contact from them
+        models.Contact.objects.create(**kwargs)
+
+        # we will test groups test seperately then take out of kwargs
+        self.assertEqual([g.uuid for g in kwargs['groups']], ['G-001', 'G-007'])
```
dpoirier: I'm getting a failure here just because the lists are in different orders.

tracpro/contacts/migrations/0017_auto_20170915_2217.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='contact',
+            name='group',
+        ),
+        migrations.AlterField(
+            model_name='contact',
+            name='groups',
+            field=models.ManyToManyField(help_text='All cohorts to which this contact belongs.', related_name='contacts', verbose_name='Cohorts', to='groups.Group'),
```
dpoirier: The convention is not to capitalize verbose names. Django will capitalize them when needed.

dpoirier:
``get_contacts_in_groups`` in the client needs some changes - I think we discussed this earlier. Right now it's likely returning multiple copies of contacts that are in multiple groups, and setting a group_uuid field on each of them based on the group fetch that copy came back in, but that's not really very useful. I'm also not sure what happens when we process those multiple copies of contacts in the sync.

Every copy of a contact that we get back from rapidpro ought to have a complete list of the groups it belongs to (but check that, please), so we should only need to keep the first copy we get of each contact. So I suggest we modify get_contacts_in_groups to return a list that only has one copy of each contact, and not to set ``group_uuid`` on them.

Then the parts of the sync code that look at group_uuid will need to accomplish their tasks some other way.  I suspect Contact.kwargs_from_temba is already doing what is needed, and we just don't need the code that looks at group_uuid at all.

dpoirier:
re: ``get_response_counts``, here's how I'd proceed.

First, I'd rename some variables to make it easier to follow what's going on. I'd suggest renaming ``qs`` to ``responses`` and ``q`` to ``response``, at least.

Next, I'd figure out, maybe looking at the original code, exactly what this is supposed to do - what's the format of the return value, and what are the callers expecting to be in there.

Then write a detailed docstring for this method explaining that.  As an example - and I'm just speculating here about the details, I haven't studied this code that carefully yet - it might say something like "This method returns a dictionary. Each key is a Group or Region object, and each associated value is the number of responses received in the last 30 days from contacts belonging to that group."

That way we are sure we understand what this method is supposed to be doing and document it for later developers.

Maybe then it'll be clearer whether we're attacking this in a valid way.

tracpro/contacts/models.py
```diff
@@ -142,7 +138,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         self._data_field_values = kwargs.pop('_data_field_values', None)
-        super(Contact, self).__init__(*args, **kwargs)
+        kwargs.pop('groups', None)
```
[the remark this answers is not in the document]
philtuskha: From contacts/utils.py part of sync_pull_contacts(), starting at line 102: 

`          # create new contact from kwargs
            new_contact = Contact(**kwargs)
            new_contact.save()

            # add all the related groups to
            new_contact.groups = kwargs['groups']`

Here, I've sent the kwargs to create a new contact, but `groups` is not a valid keyword for this __init__ function, so I pop it out of the kwargs.  and then add it into the new_contact.  I didn't really enjoy doing this :), but didn't want to mess this function up if it were being used elsewhere.  I was entertaining adding a `create` class method instead.  Somehow I thought that that would do the job better.

It seems that I maybe should pop the kwargs out on line 101 of contacts/utils.py instead, since I'm really targeting that functionality.

tracpro/contacts/tests/test_models.py
```diff
@@ -73,20 +73,19 @@ def test_create(self):
...
         self.assertEqual(contact.name, "Ann")
 
-    @skip("Skipping test_get_or_fetch_non_existing_local_contact() for now, fixing functionality for API v2.")
     def test_get_or_fetch_non_existing_local_contact(self):
         mock_contact = TembaContact.create(
             name='Mo Polls',
             uuid='C-009',
             urns=['tel:123'],
-            groups=[self.region1.uuid, self.group1.uuid],
+            groups=[self.region1, self.group1],
```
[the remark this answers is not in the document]
philtuskha: I think since this mock_contact is supposed to come from RapidPro, the idea is that groups can be, what is known to TracPro as, either a Panel (region) or a Cohort (group).  Is that more confusing or less confusing? 😄

tracpro/polls/views.py
```diff
@@ -361,7 +361,7 @@ def get_context_data(self, **kwargs):
             # Calculate all reporter group or region activity per group or region
             for group_or_region in groups_or_regions:
                 if group_by_reporter_group:
-                    responses_group = responses.filter(contact__group=group_or_region)
+                    responses_group = responses.filter(contact__groups__name=group_or_region)
```
dpoirier: ``group_or_region`` looks like it's a ``Group`` here, so querying for matches against the names of the contact's groups wouldn't match. I think we want something like ``contact__groups=group_or_region``.

tracpro/polls/views.py
```diff
@@ -376,7 +376,7 @@ def get_context_data(self, **kwargs):
 
             # Calculate all no-group or no-region activity
```
dpoirier: Just noticing that this whole little section, as it already existed, seems fishy.  We would never have a contact with no groups, because we only sync contacts that have at least one group in our selected cohorts and one group in our selected panels.

This isn't the time to go tearing into other parts of the code, but maybe we should open a ticket to look into this at some point?  It doesn't *seem* like it could be doing anything useful.

tracpro/polls/views.py
```diff
@@ -543,8 +544,8 @@ def lookup_field_label(self, context, field, default=None):
         def lookup_field_value(self, context, obj, field):
             if field == 'region':
                 return obj.contact.region
-            elif field == 'group':
-                return obj.contact.group
+            elif field == 'groups':
+                return ', '.join(group.name.encode('utf-8') for group in obj.contact.groups.all())
```
dpoirier: We might want to order these groups by name the same as we do a few lines farther down, just for consistency, but not a big deal either way.

tracpro/polls/views.py
philtuskha: +1

philtuskha:
Note: 
- Also ordered queryset by name in contacts/views.py for further consistency
- Caught a lingering comment that you had noticed before and removed
- Fixed test you had mentioned/fixed in hipchat

PR #253, opened 2017-09-22
[TD-58] Fix handling when using latest answer

dpoirier:
Need to set all answers' "value_to_use" to the latest
answer on that day, if the org is configured that way;
apparently cannot rely on charts just using the answer
from the latest response. (?)

### DIFF
--- a/tracpro/orgs_ext/forms.py
+++ b/tracpro/orgs_ext/forms.py
@@ -2,7 +2,6 @@ from __future__ import unicode_literals
 
 from dash.orgs.forms import OrgForm
 from django.contrib.auth.models import User
-from django.db.models import F
 from django.db.models.functions import Lower
 
 from temba_client.base import TembaAPIError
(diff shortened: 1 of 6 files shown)


tracpro/orgs_ext/forms.py
```diff
@@ -162,23 +161,16 @@ def save(self, *args, **kwargs):
 
         instance = super(OrgExtForm, self).save(*args, **kwargs)
```
philtuskha: I was getting the opposite behavior from what was expected when you update the orgs `how_to_handle_sameday_responses`.  I was able to add instance.save() and get the expected result, but I suspect that you could save in the if statement above (line 152).  Though, I'm not totally sure of the right convention here.

tracpro/orgs_ext/signals.py
```diff
@@ -14,3 +15,17 @@ def set_visible_data_fields(sender, instance, **kwargs):
...
+    """
+    Hook to update 'value_to_use' in numeric answers based on how_to_handle_sameday_responses.
+    Unfortunately we can't do this in the form.save() method because the admin doesn't commit
+    the save until later.
+    """
+    for answer in Answer.objects.filter(
+        question__poll__org=instance,
+        question__question_type=Question.TYPE_NUMERIC,
+    ).distinct():
+        answer.update_own_sameday_values_and_others()
```
philtuskha: 🎉

PR #254, opened 2017-09-25
Fix bug syncing twitter contacts

dpoirier:
RapidPro seems to be converting URNs that use
Twitter handles to ones using Twitter IDs, which tracpro
doesn't know about. Update the list of URN schemes that
we understand to the current list on RapidPro.  (All we
really need to know about these is a translatable label
for each one, no additional code needed on our end.)


tracpro/contacts/models.py
```diff
@@ -95,9 +99,9 @@ class Contact(models.Model):
         verbose_name=_("Full name"), max_length=128, blank=True,
         help_text=_("The name of this contact"))
     urn = models.CharField(
-        verbose_name=_("Phone/Twitter"),
+        verbose_name=_("Phone/Twitter etc."),
         max_length=255,
-        help_text=_("Phone number or Twitter handle of this contact."))
+        help_text=_("How to communicate with this contact."))
```
philtuskha: Maybe we should put some valid examples in the help text, especially for the twitter handle.  I noticed the contacts won't sync without the country code (giving the not-so-visible warning below), but won't give an error on the form if the urn is incorrect. So, it would be nice to at least give an example in the help_text.  Maybe?

dpoirier:
@philtuskha I simplified this some, but also added help for the URN formats, can you take another look?

PR #258, opened 2017-09-28
Validate contact form urns

philtuskha:


tracpro/contacts/fields.py
```diff
@@ -41,14 +44,75 @@ def render(self, name, value, attrs=None):
...
+        validate_numeric_twitter_id(numeric_id)
+    else:
+        raise ValidationError(
+            _("That is not a valid format Twitter ID.  The correct format is: "
+              "<numeric id>#<Twitter handle>"))
+
+
+def validate_phone_used(number):
+    from .models import Contact
+    contacts = Contact.objects.all()
```
dpoirier: To see if a number is in use, it's probably more efficient to just query for contacts with that phone number, e.g. ``Contact.objects.filter(urn="tel:%s" % number)``.

However, we don't really know here whether the number we're validating belongs to a contact that already exists. If so, it's okay for the number to be in use, because it's in use by the contact we're validating it for.  Typically we'd handle this by excluding the pk of the existing contact from the query, so we'll have to pass that along from the ``clean`` method if the record has a pk.  For that matter, the ``clean`` method is on the field class, and I don't know offhand whether it even has access to the record it belongs to.  You'd hope so, of course.

Actually there's another problem - we should be ignoring any punctuation in the number, since the same number could be written different ways. I'm not really sure how to make that efficient, except maybe to store another field with the canonical form of the urn, with punctuation stripped or something, and use that to check for dupes. This gets more and more messy, unfortunately.

Which leads me to wonder if it's really worth trying to avoid duplicates here, if it's going to be this much work. Do we know if RapidPro allows contacts with the same phone number?

tracpro/contacts/fields.py
```diff
@@ -41,14 +44,75 @@ def render(self, name, value, attrs=None):
...
+
+    if number in phone_numbers:
+        raise ValidationError(
+             _("This phone number is already being used by another contact."))
+
+
+def validate_phone(number):
+    try:
+        parsed = phonenumbers.parse(number)
+    except Exception as error:
```
dpoirier: Here, we should only catch the specific exception that we're expecting if the number is not valid. If anything goes wrong that we didn't expect, we'd like to find out about it and not hide it.  This case isn't quite so bad in that way, since it's still going to report an error, but not all exceptions have an ``_msg`` attribute, and it's just a good habit to get into to catch the narrowest possible set of exceptions.

tracpro/contacts/fields.py
```diff
@@ -41,14 +44,75 @@ def render(self, name, value, attrs=None):
...
+        raise ValidationError(
+            _(error._msg))
+    else:
+        reason = phonenumbers.is_possible_number_with_reason(parsed)
+        if reason == 1:
```
dpoirier: Instead of using literal numbers here, we should import names from phonenumbers and use them. That makes the code clearer and more readable, and easier to spot if we've got a mismatch between the case we're checking for and the error we're going to report.

tracpro/contacts/fields.py
```diff
@@ -58,3 +122,19 @@ def __init__(self, *args, **kwargs):
...
+            validate_twitter_handle(address)
+            return compressed_value
+
+        elif scheme == URN_SCHEME_TWITTERID:
+            validate_twitter_id(address)
+            return compressed_value
+
+        elif scheme == URN_SCHEME_TEL:
+            validate_phone(address)
+            return compressed_value
```
dpoirier: If we get here without recognizing the scheme, we should raise a validation error.  Hopefully that won't happen (our form supposedly limits them to the schemes we know about), but say we added another scheme and forgot to update this code, we'd rather immediately get an error when we try it, than mysteriously be saving "None" as the value because that's what ``clean`` returned here.

tracpro/contacts/tests/test_views.py
```diff
@@ -26,6 +26,8 @@ def test_create(self):
         # log in as an org administrator
         self.login(self.admin)
 
+        # try:
+        #     with transaction.atomic():
```
dpoirier: Left-over from trying something else?

tracpro/contacts/tests/test_views.py
```diff
@@ -112,6 +96,148 @@ def test_create_with_access(self):
...
+            'urn_0': "tel",
+            'urn_1': "2345263746",
+            'region': self.region1.pk,
+            'groups': (self.group1.pk, self.group2.pk, self.group3.pk),
+        })
+
+        self.assertEqual(response.status_code, 200)
+        form = response.context['form']
+        self.assertEqual(len(form.errors), 1, form.errors)
+        self.assertFormError(response, 'form', 'urn', ('Missing or invalid default region.'))
```
dpoirier: It seems like the error here should be the one we used earlier, "This phone number has an invalid country code."?  (which seems much clearer than this one)

tracpro/contacts/tests/test_views.py
```diff
@@ -112,6 +96,148 @@ def test_create_with_access(self):
...
+            'region': self.region1.pk,
+            'groups': (self.group1.pk, self.group2.pk, self.group3.pk),
+        })
+
+        self.assertEqual(response.status_code, 200)
+        form = response.context['form']
+        self.assertEqual(len(form.errors), 1, form.errors)
+        self.assertFormError(response, 'form', 'urn', ('This phone number is too short.'))
+
+        # phone number has already been used
```
dpoirier: wrong comment here

tracpro/contacts/fields.py
philtuskha: +1 on filtering and excluding(when Updating).  I left this in a state that 'worked' but then realized that this might be a little more involved than I have here, I will discuss below.*

Yes, I think I have a solution for the telephone number having  

* Rapidpro doesn't allow duplicate phone numbers, but this is actually for the whole org, not just the contacts that TracPro has synced based on Panels and Cohorts.  When you try to sync the new contact (that you were able to save on TracPro given that the number was unique for all TracPro contacts), the sync still fails in the background with an error message not visible to a user (error: tracpro/contacts/utils --> line 187).  I'm deciding what to do here.

tracpro/client.py
```diff
@@ -51,6 +51,20 @@ def get_contacts_in_groups(self, group_uuids, deleted=None):
...
+        indicating that the contact is being updated
+        """
+        contacts = self.get_contacts(urn=urn)
+        if len(list(contacts)) > 0 and contacts[0].uuid != uuid:
+            return True
+        elif len(list(contacts)) > 0 and contacts[0].uuid == uuid:
+            return False
+        else:
+            return False
+
```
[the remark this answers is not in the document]
philtuskha: +1 on shortening to return value of the expression.  

Otherwise, I agree that we should be able to just write `if contacts`, and started off doing so, but I get a key error from line 81 of the same page.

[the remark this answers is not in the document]
dpoirier: That sounds like a bug in our client code.

Maybe apply this change and try again?

        diff --git a/tracpro/client.py b/tracpro/client.py
        index c1b5f15..8655147 100644
        --- a/tracpro/client.py
        +++ b/tracpro/client.py
        @@ -70,3 +70,9 @@ class TracProCursorQuery(CursorQuery):

             def __getitem__(self, index):
                 return self._get_result()[index]
        +
        +    def __len__(self):
        +        return len(self._get_result())
        +
        +    def __nonzero__(self):
        +        return len(self._get_result()) != 0

tracpro/contacts/fields.py
```diff
@@ -44,70 +56,98 @@ def render(self, name, value, attrs=None):
...
+
+    if urn.startswith('twitterid'):
+        contacts = Contact.objects.filter(urn__icontains=urn.split('#')[0])
+        contacts = contacts.exclude(uuid=uuid)
+    else:
+        contacts = Contact.objects.filter(urn=urn).exclude(uuid=uuid)
+    if contacts.count() != 0:
+        return True
+    else:
+        if client.check_urn_exists(urn, uuid) is True:
```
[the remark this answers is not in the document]
philtuskha: Noted and will think to refactor as I go.  This `client.check_urn_exists(urn, uuid) is True` is explicitly in place because of testing.  Mocking, specifically, makes it so a test fails with a 200 error with an error stating that the phone number already exists, instead of a 302 as expected.  Even when a phone number does not exist.

tracpro/contacts/fields.py
```diff
@@ -58,3 +162,25 @@ def __init__(self, *args, **kwargs):
...
         return '%s:%s' % (values[0], values[1])
+
+    def clean(self, value, *args, **kwargs):
+        compressed_value = super(URNField, self).clean(value)
+        scheme, address = compressed_value.split(':', 1)
+
+        client = get_client(self.org)
+
+        if scheme == URN_SCHEME_TWITTER:
+            validate_twitter_handle(client, self.uuid, address)
```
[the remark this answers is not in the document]
philtuskha: So, it's true when you are adding a new contact, self.uuid does not exist.
But, I passed it into field from form here:

`self.fields['urn'] = URNField(label=_("Phone/Twitter"), org=org, uuid=self.instance.uuid)`

When you are updating, this does exist.

tracpro/contacts/fields.py
```diff
@@ -44,70 +56,98 @@ def render(self, name, value, attrs=None):
...
+def check_urns(client, uuid, urn):
+    '''Checking the local Contacts for a urn.
+        If none exist, checking RapidPro for a urn.
+        If none exist, returning False.'''
+    from .models import Contact
+    # RapidPro checks for the numeric id without handle
+    # we are going to mimic that here locally
+
+    if urn.startswith('twitterid'):
+        contacts = Contact.objects.filter(urn__icontains=urn.split('#')[0])
```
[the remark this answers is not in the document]
philtuskha: +1 for `urn__startswith`
 `urn.split('#')[0]` contains the string, 'twitterid'

tracpro/contacts/fields.py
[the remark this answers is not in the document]
dpoirier: Since the client is mocked, the mock object won't have a check_urn_exists method on it, and so trying to call that will just return another mock, which will be "true-ish".  If you just want to make check_urn_exists return True or False, you can do something like this first::

    client.check_urn_exists.return_value = False

and then anything calling check_urn_exists on that mock client will get back False.

If you need to test check_urn_exists itself, you'll probably need to use a separate test class that doesn't inherit from the Tracpro test class that always mocks the client. Or create a new client in your test and use that.

tracpro/contacts/fields.py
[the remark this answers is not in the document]
dpoirier: ah, okay, I had missed that.

tracpro/contacts/fields.py
```diff
@@ -63,12 +63,10 @@ def urn_exists_in_rapidpro(client, uuid, urn):
...
-
-    if uuid:
-        return len(list(contacts)) != 0 and contacts[0].uuid != uuid
-    else:
-        if len(list(contacts)) != 0:
-            return contacts[0].uuid is not None
+    try:
+        return contacts[0].uuid is not None and contacts[0].uuid != uuid
+    except:
+        return False
```
philtuskha: With an actual client, if there are no contacts associated with the passed in urn here, `contact` will be an empty list. When mocking a client, an empty contact is made (That is to say, a contact with empty parameters).  Hence, this code.

dpoirier: If for a test, we want to simulate there being no contacts, then we need to arrange for the mock client to return the same thing a real client would return in that case. Let's get together this morning and we can dig into what's going on there and sort it out.

tracpro/contacts/fields.py
```diff
@@ -41,14 +57,110 @@ def render(self, name, value, attrs=None):
...
+    else:
+        return urn
+
+
+def validate_phone(client, uuid, number):
+    try:
+        parsed = parse(number)
+    except NumberParseException as error:
+        raise ValidationError(
+            _(PHONE_PARSE_ERROR[error.error_type]))
```
dpoirier: Sorry I just noticed this: the ugettext_lazy() call (here aliased to ``_``) needs to be wrapped around the literal text that needs translating. It won't work wrapped around a variable reference or more complicated expression.

(I can explain this IRL if you want, it's too much to type here :-))

PR #266, opened 2017-10-05
Increase test coverage

dpoirier:
Coverage was already around 80%, but I was able to clean some
things up:

* Always report coverage when running tests with 'make -k', even
  if some tests fail.
* Use 'make -k' from Travis as well, rather than running different
  test commands there.
* Get Javascript linting to pass again (including deleting a
  bootstrap.js file that was completely unused)
* Add a few new tests.


dpoirier:
Now this is passing on Travis.

PR #268, opened 2017-11-27
[TD-67] Include all responses as-is in CSV downloads

dpoirier:
When we show charts of responses, we pretend each contact only
responded once to the poll each day, either taking the last response
or the sum of their responses for the day.

When downloading CSV, we instead want to include all the
responses, exactly as they were answered.

### DIFF
--- a/tracpro/polls/tests/test_views.py
+++ b/tracpro/polls/tests/test_views.py
@@ -32,17 +32,33 @@ class ResponseCRUDLTest(TracProDataTest):
 
     def setUp(self):
         super(ResponseCRUDLTest, self).setUp()
+        date0 = datetime.datetime(2014, 1, 1, 6, 59, 59, tzinfo=pytz.UTC)  # Slightly earlier than date1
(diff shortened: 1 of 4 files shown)


tracpro/polls/tests/test_views.py
```diff
@@ -127,7 +145,8 @@ def test_fetching_pollruns_csv(self):
         self.assertEqual(200, response.status_code)
         self.assertEqual('text/csv', response['Content-Type'])
 
-        rows = [row for row in csv.reader(StringIO(response.content.decode('utf-8')))]
+        rows = [[element.decode('utf-8') for element in row]
+                for row in csv.reader(StringIO(response.content.decode('utf-8')))]
```
philtuskha: `row[0]` (line 150, below) would also contain 'Cohorts' instead of 'Cohort' when changed in views.

dpoirier:
Thanks Phil, I've updated this pull request with that change.

PR #275, opened 2018-02-22
Fix bug responding to a conversation

dpoirier:
Fix a bug reported by Shane when trying to respond to a conversation.

First added a test and made sure it reproduced the problem, then fixed the problem so the test passed.

Also, make a change to double-check that Travis is giving us the postgres version we asked for, which I added before determining the real problem was that psycopg2 needed to be upgraded to fix a bug parsing postgres versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ cache:
 env:
   - DJANGO_SETTINGS_MODULE="tracpro.settings.tests"
 install:
+  - psql --version
+  - echo "select version();" | psql
   - pip install --upgrade pip
   - pip install -r requirements/tests.txt
   - npm install less
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ cache:
 env:
   - DJANGO_SETTINGS_MODULE="tracpro.settings.tests"
 install:
-  - psql --version
-  - echo "select version();" | psql
   - pip install --upgrade pip
   - pip install -r requirements/tests.txt
   - npm install less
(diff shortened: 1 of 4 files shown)


tracpro/msgs/tests/test_views.py
```diff
@@ -90,6 +90,23 @@ def test_conversation(self):
...
+        url = reverse('msgs.inboxmessage_conversation', args=[self.contact1.pk])
+
+        # a task will try to fetch the latest messages from rapidpro
+        self.mock_temba_client.get_messages.return_value = []
+
+        # Post a short message as a response.
+        response = self.url_post('unicef', url, data={'text': 'Now is the time'}, follow=True)
+        self.assertRedirects(
+            response,
+            reverse('msgs.inboxmessage_conversation', kwargs=dict(contact_id=self.contact1.pk)),
```
vkurup: This reverse call looks almost identical to the `url` value in line 95. Can we reuse it here? Or is it not really identical.

.travis.yml
```diff
@@ -12,6 +12,8 @@ cache:
 env:
   - DJANGO_SETTINGS_MODULE="tracpro.settings.tests"
 install:
+  - psql --version
+  - echo "select version();" | psql
```
vkurup: Did you want to keep these around? Fine if so... just checking to make sure it wasn't a temporary debugging step you wanted to do as a one-time.

vkurup (approved):
A couple questions, but :shipit: either way.

.travis.yml
dpoirier: I might as well clean these up, though they don't hurt anything.

tracpro/msgs/tests/test_views.py
dpoirier: I hadn't noticed that, thanks. I'll do that.

